PR #44, opened 2020-02-05
deep reify local links

### LINKED ISSUE #43: [BUG] bad behavior in npm/cli#750 test case
Set up a folder like described in https://github.com/npm/cli/issues/750

Run the `scripts/reify.js` script on it, with `--save`

1. Does not create link to `./lib` in `node_modules`.
2. symlink is absolute url, should be relative.
3. `package-lock.json` looks like this:

```json
{
  "name": "monorepo",
  "lockfileVersion": 2,
  "requires": true,
  "packages": {
    "": {
      "name": "monorepo",
      "dependencies": {
        "app": "file:./app"
      }
    },
    "app": {},
    "node_modules/app": {
      "resolved": "app",
      "link": true
    }
  },
  "dependencies": {
    "app": {
      "version": "file:app"
    }
  }
}
```

Run `scripts/reify.js --save` on it again, and now package-lock.json looks like this:

```json
{
  "name": "monorepo",
  "lockfileVersion": 2,
  "requires": true,
  "packages": {
    "": {
      "name": "monorepo",
      "dependencies": {
        "app": "file:./app"
      }
    },
    "app": {
      "resolved": "git+ssh://git@github.com/node_modules/app.git"
    },
    "node_modules/app": {
      "resolved": "app",
      "link": true
    }
  },
  "dependencies": {
    "app": {
      "version": "file:app"
    }
  }
}
```

Super weird, looks like it's interpreting the `node_modules/app` reference as a github shorthand?  That's wrong.

isaacs:
Handle link deps properly in ideal/virtual/reify

Fix: #43

This fixes the bug identified by https://github.com/npm/cli/issues/750

Link deps that point to a location underneath the root package are fully
resolved and reified.  Links external to the root path are left alone,
because they exist in a different project and presumably have their own
dep resolution being done separately.

Also, Link nodes now have their `resolved` field properly relativized to the _directory_ where the link resides, rather than against the link path itself, since that is how symlinks work.

cc: @wesleytodd I believe this may have some impacts on your explorations of workspaces.  Not sure if it's helpful or harmful, but most likely something to be aware of.


wesleytodd:
> cc: @wesleytodd I believe this may have some impacts on your explorations of workspaces. Not sure if it's helpful or harmful, but most likely something to be aware of.

I was directly doing this myself.  So this is an improvement IMO.  The one thing I would like to see is an option to also reify the external links.  Your presumption is only valid because that is how other tools have treated it, but as I mentioned to @ruyadorno, that is because they were hacking on top of the filesystem based resolution.  If you have something like [import maps](https://github.com/nodejs/node/issues/49443) or yarn style PnP you don't have this requirement anymore.  I think there are many use cases for having a one step process to reify across many unrelated "projects" if you need to work on them together.

isaacs:
@wesleytodd Thanks.  I can certainly imagine cases where it would be unwelcome (or just impossible), and many others where it would be super convenient.  (For myself, most of the times that I _didn't_ want link deps messed with, it was because npm did the wrong thing, but Arborist doesn't, so it's usually going to be fine?)

In any event, it's just a boolean check at one point, so it's easy enough to add a flag for.  I guess then the question would be whether this should be opt-in or opt-out.  Any preference?

wesleytodd:
> it was because npm did the wrong thing, but Arborist doesn't, so it's usually going to be fine?

Yeah that logic makes sense, but I am a big fan of it "doing the right thing!

>  I guess then the question would be whether this should be opt-in or opt-out. Any preference?

Opt in I think.  My guess is that it will take time and usage for folks to see how this all works, and it is a bigger change if it is the default behavior.   My use case is such that I am happy to opt-in, and I imagine others who dig in like me will feel the same.

PR #50, opened 2020-02-24
feat: npm workspaces

ruyadorno:
## Overview

Introduces support to **workspaces**; adding ability to build an ideal tree that links defined `workspaces`, storing and reading lockfiles that contains `workspaces` info and also reifying installation trees properly symlinking nested `workspaces` into place.

## :pencil2: Changes
- `lib/arborist/build-ideal-tree.js` Creates trees that take into account `workspaces` definitions
- `lib/arborist/load-virtual.js` Added ability to read `workspaces` definitions
- `lib/edge.js` Added check that a `workspace` edges are symlinks
- `lib/node.js` Added `workspaces` setter that will properly create the `edgesOut` pointing to nested workspaces
- `lib/shrinkwrap.js` Add `workspaces` info to lockfile

## :link: References
- https://github.com/npm/rfcs/blob/ea2d3024e6e149cd8c6366ed18373c9a566b1124/accepted/0026-workspaces.md
- https://www.npmjs.com/package/@npmcli/map-workspaces
- https://github.com/npm/rfcs/pull/103


lib/arborist/load-virtual.js
```diff
@@ -15,6 +16,7 @@ const resolveLinks = Symbol('resolveLinks')
 const assignParentage = Symbol('assignParentage')
 const loadNode = Symbol('loadVirtualNode')
 const loadLink = Symbol('loadVirtualLink')
+const loadWorkspaces = Symbol('loadWorkspaces')
```
isaacs, line 19: I'm just noticing this now, not relevant to this PR, but we should make loadVirtual underscore its symbols like the other mixin classes.

lib/arborist/build-ideal-tree.js
```diff
@@ -234,9 +237,19 @@ module.exports = cls => class IdealTreeBuilder extends Tracker(Virtual(Actual(cl
       optional: false,
       global: this[_global],
       legacyPeerDeps: this.legacyPeerDeps,
+      hasWorkspaces: !!pkg.workspaces,
```
isaacs, line 240: I don't see where this field is used, is it a leftover from something that changed?

ruyadorno: oh yeah def a leftover, will clean it up 😊

PR #52, opened 2020-03-12
mikemimik/feature/expose diff

mikemimik:
# What / Why

Expose the tree diff `Diff` node, in order for the `cli` to be able to print the results of an arborist action.

## References

* https://github.com/npm/cli/issues/995
* https://github.com/npm/cli/pull/1017

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -98,6 +98,7 @@ module.exports = cls => class Reifier extends Ideal(cls) {
       scriptShell,
       savePrefix = '^',
       binLinks = true,
+      save = true,
     } = options
 
     if (!binLinks)
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -98,7 +98,6 @@ module.exports = cls => class Reifier extends Ideal(cls) {
       scriptShell,
       savePrefix = '^',
       binLinks = true,
-      save = true,
     } = options
 
     if (!binLinks)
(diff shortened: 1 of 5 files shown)


lib/arborist/reify.js
```diff
@@ -99,6 +98,7 @@ module.exports = cls => class Reifier extends Ideal(cls) {
       scriptShell,
       savePrefix = '^',
       binLinks = true,
+      save = true,
```
isaacs: As much as it's good to be explicit about what we're pulling off the options object at each stage, since this variable isn't actually used in this scope, I think it'd be better to remove this line.

test/arborist/build-ideal-tree.js
```diff
@@ -205,6 +205,15 @@ t.test('dedupe example - deduped', t => {
   return t.resolveMatchSnapshot(printIdeal(path), 'dedupe testing')
 })
 
+t.test('expose explicitRequest', async t => {
+  const path = resolve(__dirname, '../fixtures/simple')
+  const arb = new Arborist({ path })
+  const tree = await arb.buildIdealTree({ add: [ 'abbrev' ] })
+  t.ok(arb.explicitRequests, 'exposes the explicit request')
```
isaacs: Should test type and value, too.  Something like this: `t.strictSame(arb.explicitRequests, new Set(['abbrev']))`

lib/arborist/reify.js
mikemimik: Yeah, I concur with that. I'll pull it out.

test/arborist/build-ideal-tree.js
mikemimik: Totally, will add this!

PR #58, opened 2020-03-30
track timing via process time/timeEnd events

### LINKED ISSUE #55: [FEATURE] timing
It would be nice if reify() (and, for that matter, other functions as well) would track how long things take.

Call `process.emit('time', name)` when a thing starts, and `process.emit('timeEnd', name)` when they end.

isaacs:
Emit a 'time' event for various things, and then a timeEnd event when that
thing is completed.  This is picked up by the CLI in its logger when the
loglevel is set to 'timing', but can also be consumed by any module using
arborist in a generic way.

Close #55

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -119,6 +119,8 @@ module.exports = cls => class IdealTreeBuilder extends Tracker(Virtual(Actual(cl
     if (this.idealTree)
       return Promise.resolve(this.idealTree)
 
+    process.emit('time', 'idealTree')
+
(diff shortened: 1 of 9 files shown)


isaacs:
Note: removed the timers from `lib/arborist/load-actual.js`, because that function is called in a few places during reification.  If there are multiple deps with bundled dependencies, then it's possible for there to be multiple `loadActual` calls happening at once, and it makes the timers get screwy.  Also, it's wicked fast, and we're tracking the `initTrees` portion of `reify()`, so it's pretty much covered.

Issue #54, opened 2020-03-27
[BUG] audit


isaacs:
# quick audit on reify

Arborist should construct the audit payload and submit the quick audit for all reifications (unless `audit: false` is in the options) and have a function to perform a full audit.

# full audit without fix

Load the actual tree, and submit to the full audit endpoint.  Return the audit report object.

# full audit with fix

Submit the full audit, then perform the necessary remediations as specified.

- Copy actualTree to this.idealTree
- Make remediation changes to this.idealTree
- Call this.reify() (triggers another quick audit, which should pass)

isaacs:
Quick audit on reify: 8ff0dc38f62051d0f378c72fbcd267e67fff85da

PR #59, opened 2020-03-31
implement `npm audit` functionality

isaacs:
good to review now

#54


isaacs:
Just lacking unit tests for `buildIdealTree` now.  Those code paths are only exercised via the `arb.audit({ fix: true })` call to `this.reify()` which calls `this.buildIdealTree()` and uses `this.auditReport` to know which versions to try to avoid.

isaacs:
Ok, this is ready to go!

tap-snapshots/test-arborist-reify.js-TAP.test.js
```diff
@@ -13521,23 +13521,55 @@ Node {
...
         },
       },
+      "edgesOut": Map {
+        "@ruyadorno/package-with-added-bin" => Edge {
+          "name": "@ruyadorno/package-with-added-bin",
+          "spec": "^2.0.0",
+          "to": "b/node_modules/@ruyadorno/package-with-added-bin",
+          "type": "prod",
+        },
+      },
```
ruyadorno, line 13572: why are these new `edgesIn` and `edgesOut` getting added here?

isaacs: Because I had to make `loadVirtual` read the package.json on linked packages, or else their dependencies don't get loaded, and we can't calculate the audit report properly.

This is sort of a deeper version of the fix we made in buildIdealTree to fix the issue you found earlier with bin files.  Those edges _should_ be there, since they're a legitimate dependency in the graph, but weren't found, because the legacy shrinkwraps don't store dependency relationships for link targets.

isaacs:
Ok, now this is really truly ready for review.

Added the toJSON method so that we can serialize the object for consumption by npm-audit-report.

tap-snapshots/test-arborist-reify.js-TAP.test.js
ruyadorno: oohh makes sense! 👍

PR #63, opened 2020-04-22
test: asymmetrical bin across versions repro

ruyadorno:
Added a test that reproduces a problem when installing transitive deps
that contains multiple versions in which its `bin` package.json
declaration is not symmetrical across deduped versions.

When a lockfile v1 is available the installation will fail to produce
`node_modules/.bin` symlinks. A similar install setup but with an absent
lockfile will yield a functioning bin folder.

/cc @isaacs


isaacs:
Ooh, this is a good find.  There are two bugs here.

I think the "added a bin in a subsequent version" aspect is a bit of a red herring.  It's not reifying the `@ruyadorno/package-with-added-bin` dep at all, or even diffing the actual and virtual versions of it.

When `loadVirtual` loads a lockfile with links in it, it creates a `Link` node, and attempts to set the `target` to the relevant node within the lockfile.

However, legacy lockfiles didn't have a separate entry for a link target, and to make matters even weirder, would put the link target's dependencies under the link entry's `dependencies` block.  So the virtual tree has a nonsense entry of `node_modules/a/node_modules/@ruyadorno/package-with-added-bin` which _should_ be `a/node_modules/@ruyadorno/package-with-added-bin`.  (That's not really a bug that's relevant to this issue afaict, but it's just weird, and probably causes other problems.)

Since the `a` and `b` target nodes weren't ever added to the list of nodes that the virtualTree knows about, they didn't get assigned an `fsParent` to the root node.  And thus when calculating the diff, it didn't see those nodes in `root.fsChildren`, so they just got completely ignored by reify.

That's the first bug.  The second bug is that, because `bin` isn't stored in legacy lockfiles, and we use the lockfile entry as the effective package.json in virtual tree loading, we need to re-read the `bin` entry along with os/cpu/engines after unpacking and before calling `binLinks`.

isaacs:
Oh, scratch part of that, it _was_ unpacking, but not doing the rest of the reifying, because the effective location of the node switches from `node_modules/a/node_modules/..` to `a/node_modules/...`.  So the bin never even _tries_ to get created.

So, three bugs, and that one in the middle that I thought wasn't related _was_ related, just fixed by making the link target a thing that the virtual tree tracks properly.

PR #64, opened 2020-04-22
get fsChildren/bins properly from legacy lockfiles 

isaacs:
Land and fix #63


ruyadorno:
awesome! does it fixes all issues you listed in #63 comments?

isaacs:
Yup!

ruyadorno:
alright ✅ rebased and tested again the original project from which I first noticed the problem and I can confirm it works 👍

Issue #72, opened 2020-05-08
[FEATURE] loadActual() should accept {root} option


isaacs:
In `loadVirtual` we allow setting a `{root}` option, which uses the supplied node as the root of the tree, instead of creating a new one from the shrinkwrap read at this.path.  This is nice when we already have the root node, or if we need to extract a package in a temp directory to read its shrinkwrap, but want to put all the nodes on a node somewhere else (maybe not on disk yet).

We do `loadActual` in a similar way to read bundleDependencies, but have to manually copy over all the children after loading.  It'd be nice to skip that step and just build them out on the node from the start.

isaacs:
Hm, doing this in the same way that loadActual does is a bit more tricky.  There are a lot of cases in `loadActual` where we assume that `node.path` and `node.realpath` are the actual paths on the filesystem.

It'd probably be better to run the loadActual process with the root node we load, and then re-root the child/fsChild nodes once loaded.  It involves a bunch of extra tree-walking to update paths, but it's a lot less bookkeeping along the way.

isaacs:
Fixed by #77.

PR #73, opened 2020-05-09
Add {complete:true} buildIdealTree option to load sw/bundles

isaacs:
Needed for implementing 'npm install --package-lock-only'

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -3,6 +3,7 @@
 const rpj = require('read-package-json-fast')
 const npa = require('npm-package-arg')
 const pacote = require('pacote')
+const cacache = require('cacache')
 const semver = require('semver')
 const pickManifest = require('npm-pick-manifest')
 const mapWorkspaces = require('@npmcli/map-workspaces')
@@ -15,6 +16,8 @@ const Link = require('../link.js')
 const addRmPkgDeps = require('../add-rm-pkg-deps.js')
 const gatherDepSet = require('../gather-dep-set.js')
 const optionalSet = require('../optional-set.js')
+
+const loadBundle = require('../load-bundle-deps.js')
 
 // enum of return values for canPlaceDep.
 // No, this is a conflict, you may not put that package here
@@ -29,6 +32,7 @@ const REPLACE = Symbol('REPLACE')
 const {resolve} = require('path')
 const relpath = require('../relpath.js')
 
+const _complete = Symbol('complete')
 const _depsSeen = Symbol('depsSeen')
 const _depsQueue = Symbol('depsQueue')
 const _currentDep = Symbol('currentDep')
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "@npmcli/name-from-folder": "^1.0.1",
     "@npmcli/run-script": "^1.3.1",
     "bin-links": "^2.1.2",
+    "cacache": "^15.0.3",
```
ruyadorno, line 11: question: why cacache is only getting added now?

lib/arborist/build-ideal-tree.js
```diff
@@ -464,13 +469,38 @@ module.exports = cls => class IdealTreeBuilder extends Tracker(Virtual(Actual(cl
...
+      const hasBundle = bd && Array.isArray(bd) && bd.length
+      const { hasShrinkwrap } = node
+      if (hasBundle || hasShrinkwrap) {
+        const opt = { ...this.options }
+        await cacache.tmp.withTmp(this.cache, opt, async path => {
+          await pacote.extract(node.resolved, path, opt)
+
+          if (hasShrinkwrap)
+            await new Arborist({ ...this.options, path })
+              .loadVirtual({ root: node })
```
ruyadorno, line 496: wow are these composing install trees? 🤩 is it that easy? it gives me some workspaces ideas...

package.json
isaacs: Until now, Arborist didn't use cacache directly!  But cracking open tarballs requires a temp directory to extract into, and it just makes sense to use the temp directory that cacache provides, since the rest of npm does.

isaacs:
> Where did arborist landed with regards to bundledDependencies name variation? is it not supported? or is it just converted in some abstraction i might have missed?

The canonical name is `bundleDependencies`, no `dD`.  This is enforced by `read-package-json-fast` [here](https://github.com/npm/read-package-json-fast/blob/master/index.js#L42-L61), and by `read-package-json` [here](https://github.com/npm/read-package-json/blob/latest/read-json.js#L328-L353).  It occurs to me that we _don't_ check for that in the manifests we get from the registry via pacote, which I'll go add now.  (Since it's been enforced at publish time for many years now, it's probably fine, but still, good to be extra careful.)

lib/arborist/build-ideal-tree.js
isaacs: Yep!  https://github.com/npm/arborist/issues/72 is an open issue to add a similar facility to `loadActual()`, which will make the loading of bundle deps a bit easier.

lib/arborist/index.js
```diff
@@ -34,7 +35,9 @@ class Arborist extends Auditor(require('events')) {
       nodeVersion: process.version,
       ...options,
       path: options.path || '.',
+      cache: options.cache || `${homedir()}/.npm/_cacache`,
```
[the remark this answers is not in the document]
isaacs: Hmm... really depends if we want that to be a "cacache default" vs an "npm default".  It's kind of nice for cacache to just be agnostic about any npm-specific stuff, and `~/.npm/_cacache` definitely feels npm-specific.

PR #76, opened 2020-05-14
fix: avoid throwing type erros on inventory.add

ruyadorno:
Trying to load mock malformed testing data from the cli resulted in an unexpected `TypeError`. The issue can be reproduced by having a falsy funding property in a package.json, as seen here: https://github.com/npm/cli/blob/abdf52879fcf0e0f534ad977931f6935f5d1dce3/test/tap/fund.js#L81

This commit fixes the problem by avoiding trying to retrieve the `type`and `url` properties when having a `null` object reference.

While I considered the alternative of throwing a more descriptive error instead of having `TypeError: Cannot read property 'url' of null` - I think in this case not blowing up is a much more convenient and reasonable behavior.


isaacs:
LGTM!

PR #81, opened 2020-05-19
fix: Handle tarball dependencies properly

isaacs:
Stacks on #80, land that first.

For 'file:' deps (whether directory or tarball type), Node.resolved is
always absolute.  It was initially set up to be relative to the node
path, because it had to be stashed back to package.json.  But, then we
got rid of that approach in favor of a more comprehensive lockfile, and
so the added complexity of constantly re-evaluating the resolved value
as the node moves around serves no purpose.

This relative linking also _wasn't_ being handled properly in tarball
dependencies, resulting in attempting to fetch a tarball from the wrong
location, relative to the dependent's directory in node_modules, rather
than where it should be found.  Since we never store Node.resolved
anywhere, there's no need for it to be portable, and the portability had
a bug anyway.

*Shrinkwrap* files *do* have to be portable, so the paths get
relativized (and resolved) going into and out of lockfiles.  But that's
much simpler.

Tarball dependencies are more consistently handled, and now that the
Node.resolved value is absolute and predictable, it's easier to get that
right.

Tarball dependencies are properly regarded as a valid dependency
resolution, provided that they come from the same target file.  It's
still a bit of a crapshoot when we get legacy install trees, but at
least now we're not throwing out package._resolved when we load the
manifest from pacote directly.

The added test case installs a tarball dependency which depends on two
other tarballs: 1 in its own containing folder, and another outside of
its folder.  (This exact situation was a thorn in npm's side for most of
6.x, and might still be in some cases.)


tap-snapshots/test-arborist-build-ideal-tree.js-TAP.test.js
```diff
@@ -29131,6 +29131,77 @@ Node {
...
+      "to": "node_modules/@isaacs/testing-file-transitive-a",
+      "type": "prod",
+    },
+  },
+  "location": "",
+  "name": "tarball-dependencies",
+  "resolved": null,
+}
+`
+
```
ruyadorno, line 29204: just curious about the resolved `spec` values that varies across this snapshot, some of them point at `file:testing-file-transitive*` others to `file:../testing-file-transitive*`, some other to `file:./c/`

took a quick look at the fixtures and while the `file:./c/` path seems to make sense since that's a nested folder - I'm quite confused to why we then have `file:testing-file-transitive-a/` and `file:../testing-file-transitive-b/`, it seems to me it would have to be either `file:./` (for transitive-a) and then `file:../testing-file-transitive-b/` or `file:../testing-file-transitive-a/` and `file:../testing-file-transitive-b/`

isaacs: The reason for the `../` there is that the edges are resolved relative to the folder where the tarball is found.

So, the root has:

```
+-- testing-transitive-a/isaacs-testing-transitive-a-1.0.0.tgz
+-- testing-transitive-b/isaacs-testing-transitive-b-1.0.0.tgz
```

The `testing-transitive-a` package's package.json has a dep on `../testing-transitive-b/isaacs-testing-transitive-b-1.0.0.tgz`.  Ie, a reference from the `a` package to the `b` package which is its peer.

When we resolve `file` spec edges on `file`-resolved packages, the resulting target is the resolution against the dirname of the requiring module's resolved value.

isaacs: That is, the root depends on `file:testing-file-transitive-a/isaacs-testing-file-transitive-a-1.0.0.tgz`, and then `a` depends on the file at `./testing-file-transitive-b/isaacs-testing-file-transitive-b-1.0.0.tgz`, which from a's pov is `../testing-file-transitive-b/isaacs-testing-file-transitive-b-1.0.0.tgz`.

tap-snapshots/test-arborist-load-actual.js-TAP.test.js
```diff
@@ -1967,8 +1967,7 @@ Node {
       "ghtgz": Object {
         "extraneous": true,
         "integrity": "sha512-yowslMd9y/lGBCDVO0RwZoXRK5X0zMsf6XECM6DdeqN7qwVnFQ6IAwJai7BD4mVe1xOdWWqWNkuzyuStvSBnHw==",
-        "resolved": "https://codeload.github.com/isaacs/abbrev-js/tar.gz/a9ee72ebc8fe3975f1b0c7aeb3a8f2a806a432eb",
-        "version": "npm:abbrev@1.1.1",
+        "version": "https://codeload.github.com/isaacs/abbrev-js/tar.gz/a9ee72ebc8fe3975f1b0c7aeb3a8f2a806a432eb",
```
isaacs, line 1970: Interesting side effect of the changes in this: there was previously a bug where a remote tarball dep would be saved incorrectly as an alias dep.

tap-snapshots/test-arborist-build-ideal-tree.js-TAP.test.js
ruyadorno: I see! I think one piece of information I skipped and realized later is that these `spec` values are from edges, so in reality it's just whatever was declared there in this fixture deps 😊

PR #87, opened 2020-05-26
feat: dedupe

ruyadorno:
Add `arborist.dedupe()` - wrapping up the work from @claudiahdz 🥳


lib/arborist/build-ideal-tree.js
```diff
@@ -693,9 +695,11 @@ module.exports = cls => class IdealTreeBuilder extends cls {
...
   // place dep, requested by node, to satisfy edge
   [_placeDep] (dep, node, edge, peerEntryEdge = null) {
-    if (edge.to && !edge.error && !this[_updateNames].includes(edge.name) &&
-        !this[_isVulnerable](edge.to))
-      return []
+    if (edge.to &&
+      !edge.error &&
+      !this[_updateNames].includes(edge.name) &&
+      !this[_isVulnerable](edge.to) &&
+      !this[_dedupeList].has(edge.to.path)) return []
```
isaacs, line 702: style nit, should be:

```js

if (edge.to &&
    !edge.error &&
    !this[_updateNames].includes(edge.name) &&
    !this[_isVulnerable](edge.to) &&
    !this[_dedupeList].has(edge.to.path))
  return []
```

- Double-indent for line-breaks within a `()`
- Always a line-break for single-line conditionals

lib/arborist/trimmer.js
```diff
@@ -0,0 +1,35 @@
...
+  async dedupe () {
+    let tree
+    try {
+      tree = await this.loadVirtual()
+    } catch {
+      tree = await this.loadActual()
+    }
+    for (const node of tree.inventory.values()) {
+      const nodes = tree.inventory.query('name', node.name)
+      if (nodes.size === 1) continue
```
isaacs, line 22: same style nit as above.  Should be:

```js
if (nodes.size === 1)
  continue
```

lib/arborist/trimmer.js
```diff
@@ -0,0 +1,35 @@
...
+
+      for (const node of nodes) {
+        this[_dedupeList].add(node.path)
+        for (const edge of node.edgesIn) {
+          this[_depsQueue].push(edge.from)
+        }
+      }
+    }
+    return this.reify({
+      preferDedupe: true
```
isaacs, line 32: Need to include the rest of the options as well, so we can pass other things to dedupe/reify/buildIdealTree.

So, `async dedupe (options = {}) {` at the start of this function, and then here, do:

```js
return this.reify({
  ...options,
  preferDedupe: true,
})
```

(Also, trailing comma.  Sorry, I'm a pedantic linter today, I guess.)

ruyadorno: ah 😂 see, the value of a linter

PR #94, opened 2020-05-29
Add dryRun config to not touch the file system

### LINKED ISSUE #46: [BUG] Configs needing to be added/respected
- binLinks - boolean, default true, whether to link bins
- rebuildBundle - boolean, default false, rebuild bundled deps
- packageLock - boolean, default true, respect/update the package-lock/shrinkwrap file
- packageLockOnly - boolean, default false, don't reify anything, just update the lockfile
- globalStyle - boolean, install as if it's a top-level global package (ie, don't resolve any deps above the top level, except peer deps)

isaacs:
Last missing config!

Fix #46 

Difference between dryRun and packageLockOnly is that we calculate a diff, so that `npm install --dry-run` can output the changes that _would_ be made, and we don't submit an audit check.


isaacs:
Other commits added are fixing issues found in code review 2020-05-28.

isaacs:
Looks like the macOS issue on GitHub Actions is resolved.

PR #97, opened 2020-05-30
Resolve infinite dependency nesting with a symlink

### LINKED ISSUE #96: [BUG] detect and handle pathological dependency nesting scenario
https://www.npmjs.com/package/@isaacs/pathological-dep-nesting-a

npm v2 behavior: fail the install
npm v3-6 behavior: spin forever
yarn v1 behavior: use all memory and crash

We can do better.

isaacs:
Fix #96

In a case like http://npm.im/@isaacs/pathological-dep-nesting-a we have
a dependency cycle that would nest infinitely:

```
a@1 -> b@1
a@2 -> b@2
b@1 -> a@2
b@2 -> a@1
```

When trying to reify this, we have to nest another a@2 under b@1,
because it would otherwise resolve to a@1.  Then that a@2 needs a b@2,
which has to nest or otherwise resolve to b@1, and the cycle goes on.

The actual (deduped) reified tree looks like:

```
root
+-- a@1
+-- b@1 (deduped from a@1's dep)
    +-- a@2 (overriding a@1 one level up)
    +-- b@2 (deduped from a@2's dep)
        +-- a@1 (overriding a@2 from one level up)
        +-- b@1 (deduped from a@1's dep)
            +-- a@2 (loop forever...)
```

With this commit, the loop is cut by detecting in buildIdealTree that a
node being added to resolve a problemEdge is satisfying an edge that
already has an identical node in its ancestor set.

Ie, in this example, when we get to the second a@1, we need to place b@1
to satisfy its dependency, or else it will resolve to b@2 (its parent).
Since the edge is a problem, we know that the nearest resolution is
_not_ a match (or else we wouldn't be bothering to add it).  Thus, any
node in the ancestry that matchces the node we're attempting to add to
the tree is certainly going to cause an infinite regress, because its
presence has caused it to need to load another identical copy of itself.

While we could throw in this situation (and it'd certainly be better
than spinning forever trying to build an infinitely large tree), we can
do better, by instead creating a symbolic link to the matching node
further up the tree.  This is almost as efficient as throwing, and
produces a correct dependency resolution.

Thus, instead of growing infinitely, the resulting tree looks like:

```
root
+-- a@1
+-- b@1
    +-- a@2
    +-- b@2
        +-- a@1
        +-- b@1 (symlink to root/node_modules/b)
```

When using legacyBundling, the tree would look like:

```
root
+-- a@1
    +-- b@1
        +-- a@2
            +-- b@2
                +-- a@1 (symlink to root/node_modules/a)
```

Which, surprisingly, is 1 less package on disk, and perhaps the only
scenario where legacyBundling produces a _more_ deduplicated tree than
the default.

We could someday explore a way to detect all matching nodes in the tree
and replace them with symlinks, as a sort of "deduplicate more", but
since that's considerably more work and may have some unforeseen
consequences, it's left as an area for future exploration.


isaacs:
(Footnote: the branch name is a reference to Infinite Jest, an impossibly convoluted novel that, despite having sold a lot of copies and being a mark of status among certain intellectual circles, I remain convinced no one[1] has actually read[2], given its impenetrable density of reference, much like this commit.)

1: myself included
2: despite their[3] aspirations to do so
3: my

lib/node.js
```diff
@@ -502,6 +502,40 @@ class Node {
...
+    if (node.name !== this.name)
+      return false
+
+    // if they're links, they match if the targets match
+    if (this.isLink)
+      return node.isLink && this.target.matches(node.target)
+
+    // if they're two root nodes, they're different if the paths differ
+    if (this.isRoot && node.isRoot)
+      return this.path === node.path
```
ruyadorno, line 521: shouldn't it also check for `realpath` in order to cover the root-is-a-link usecase?

lib/node.js
```diff
@@ -502,6 +502,40 @@ class Node {
...
+    if (this.integrity && node.integrity)
+      return this.integrity === node.integrity
+
+    // if no integrity, check resolved
+    if (this.resolved && node.resolved)
+      return this.resolved === node.resolved
+
+    // if no resolved, check both package name and version
+    // otherwise, conclude that they are different things
+    return this.package.name && node.package.name &&
```
ruyadorno, line 533: question: when accessing `this.package` properties is there any need to be defensive about it? e.g: `this.package &&` - or is it always sure to have at least an empty object?

isaacs: It's always at least an empty object.

lib/node.js
isaacs: Hmm... probably right.

isaacs: Oh!  No, not relevant, because the `node.isLink` check above ensures that they'd only be a match if they're both links and the targets match.

PR #107, opened 2020-07-17
test: fix links-all-over fixture

ruyadorno:
`links-all-over` fixtures had a nested malformed `package.json` file, this change only fixes that and updates related snapshots.


isaacs:
LGTM.  Land at will.

PR #118, opened 2020-09-02
Refactor and explain ERESOLVE conflicts

isaacs:
When a peerDep cannot be resolved, this changes a few things.

First, the Node.explain() method is used to explain what was present
that conflicted, why it's there, what the conflict (or peer conflict)
was that we couldn't install, and why THAT node was trying to be added.
With some prettying up and conversion to human-intelligible language,
this should provide the user with enough information to make an informed
decision about what to do.

Second, we override the peer dep with a "best guess" as to what ought to
go in that location if --force is enabled.  If the target location is
the first place we're trying to place the peerDep (ie, it is the parent
of the node that has the peer dependency), and it has a direct
dependency on the node in question, then we use the parent's dependency
spec instead.

Lastly, if a _set_ of peer dependencies can be moved deeper into the
tree in order to avoid a conflict, we now detect that and do that as
well.

The "fixWithForce" boolean on the explanation object can be used to
suggest that the user MAY retry the command with --force in order to
resolve the situation.  Otherwise, --legacy-peer-deps is the other quick
fix.

With these changes in place, gatsby, reakit, doctokit, and several other
real-world projects that depend on nested conflicting peerDependencies
can be installed, albeit requiring the use of the --force flag.


lib/node.js
```diff
@@ -299,6 +310,90 @@ class Node {
...
+
+  [_explain] (edge, seen) {
+    if (this.isRoot && !this.sourceReference) {
+      return {
+        location: this.path
+      }
+    }
+
+    const why = {
+      name: this.isRoot ? this.package.name : this.name,
```
ruyadorno, line 330: I think we should make a getter for this and never have to care about it again 😂 :
```
get realname() {
  return this.isRoot ? this.package.name : this.name
}
```

lib/node.js
```diff
@@ -299,6 +310,90 @@ class Node {
...
+      version: this.package.version,
+    }
+    if (this.errors.length || !this.package.name && !this.package.version) {
+      why.errors = this.errors.length ? this.errors : [
+        new Error('invalid package: lacks name and/or version')
+      ]
```
ruyadorno, line 336: was this supposed to be `this.errors.length || !this.package.name || !this.package.version` ?

otherwise the error message and be just `invalid package: lacks name and version`

lib/node.js
```diff
@@ -299,6 +310,90 @@ class Node {
...
+      why.package = this.package
+    }
+
+    if (this.root.sourceReference) {
+      const {name, version} = this.root.package
```
ruyadorno, line 341: Is there value in using separated `name` and `version` instead of `pkgid` for `whileInstalling`?

lib/peer-set.js
```diff
@@ -0,0 +1,25 @@
...
+// this to get the set of all deps that have to be checked to ensure
+// nothing is locking them into the current location.
+//
+// this is different in its semantics from an "optional set" (ie, the nodes
+// that should be removed if an optional dep fails), because in this case,
+// we specifically intend to include deps in the peer set that have
+// dependants outside the set.
+const peerSet = node => {
+  const set = new Set([node])
+  for (const node of set) {
```
ruyadorno, line 12: ah curious, can't it just get rid of this initial `for` loop and just use the provided `node` param directly? or maybe it was meant to accept arrays in the param and it should use `const set = new Set([...node])` instead?

test/fixtures/registry-mocks/content/isaacs/testing-peer-dep-conflict-chain-dd.json
```diff
@@ -0,0 +1,3 @@
+{
+  "error": "Not found"
+}
```
ruyadorno, line 3: ah, I don't find any other ref to `testing-peer-dep-conflict-chain-dd` - is this intentional?

lib/arborist/build-ideal-tree.js
```diff
@@ -810,23 +831,41 @@ This is a one-time fix-up, please be patient...
     }
 
     if (!target) {
-      const current = node.resolve(edge.name)
-      throw Object.assign(new Error('unable to resolve dependency tree'), {
+      const curNode = node.resolve(edge.name)
+      const pc = this[_peerConflict] || { peerConflict: false, current: null }
```
ruyadorno, line 835: is this property `peerConflict` used anywhere?

```suggestion
      const pc = this[_peerConflict] || { current: null }
```

ruyadorno (commented):
A few comments to read through, if all of them are just expected behavior then it LGTM 👍

lib/node.js
isaacs: Hmm, I think you're right.  It was changed from `!(this.package.name && this.package.version)`, and I did not take heed of the good Mr. DeMorgan.

lib/node.js
isaacs: Occasionally the `node.package` won't have that, if it's being set by something other than reading a `package.json`, I believe?  I seem to recall that showing up in some cases I was poking at, but didn't dig into it.

lib/peer-set.js
isaacs: The set is initialized to `new Set([node])` but then we add to the set in the loop, so it gets longer.

Ie, we start with 1 node, then if it has peer deps, we add those to the set.  Using a set rather than building up an array so that we get deduplication for free.

test/fixtures/registry-mocks/content/isaacs/testing-peer-dep-conflict-chain-dd.json
isaacs: Arg, it looks like I tried to fetch these before they switched over to `access: public` on the registry.  There seems to be a delay where they're private for a moment.  I'll update.

But yes, there's nothing using these at the moment.

lib/arborist/build-ideal-tree.js
isaacs: Good catch, that should be `{ peer: null, current: null }`, not `peerConflict: false`.

test/fixtures/registry-mocks/content/isaacs/testing-peer-dep-conflict-chain-dd.json
isaacs: Oh, my mistake, no, those ended up living at ii, jj, ...

Removed.

ruyadorno (approved):
👍

PR #119, opened 2020-09-02
Refresh dep flags in virtualTree, prune extraneous from ideal

isaacs:
This fixes a situation where the root node's package.json is manually
edited, resulting in a package-lock.json which contains extraneous
dependencies.  If we don't mark those as extraneous, we end up reifying
them back to disk, even though they should be pruned.

This performs a fresh calcDepFlags step only in the case where the root
node's package does not match the edges defined in the package-lock.json,
to handle these cases thoroughly and completely, but only in when it's
truly necessary to do the full tree-walk.

cc: @d3lm

### DIFF
--- a/lib/arborist/load-virtual.js
+++ b/lib/arborist/load-virtual.js
@@ -10,6 +10,7 @@ const Shrinkwrap = require('../shrinkwrap.js')
 const Node = require('../node.js')
 const Link = require('../link.js')
 const relpath = require('../relpath.js')
+const calcDepFlags = require('../calc-dep-flags.js')
 const rpj = require('read-package-json-fast')
 
 const loadFromShrinkwrap = Symbol('loadFromShrinkwrap')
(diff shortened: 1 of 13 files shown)


lib/arborist/load-virtual.js
```diff
@@ -71,14 +79,86 @@ module.exports = cls => class VirtualLoader extends cls {
...
+      node.dev = true
+      node.optional = true
+      node.devOptional = true
+      node.peer = true
+    }
+    calcDepFlags(this.virtualTree, true)
+  }
+
+  // check the lockfile deps, and see if they match.  if they do not
+  // then we have to reset dep flags at the end.  for example, if the
```
d3lm, line 106: Maybe we can fix this comment and capitalize  `For`, and remove the extra space in front.

lib/arborist/load-virtual.js
```diff
@@ -71,14 +79,86 @@ module.exports = cls => class VirtualLoader extends cls {
...
+  // user manually edits their package.json file, then we need to know
+  // that the idealTree is no longer entirely trustworthy.
+  [checkRootEdges] (s, root) {
```
d3lm, line 109: What is `s`?

d3lm:
This is looking good! Thanks a lot @isaacs for getting this implemented so quickly. Good job 👏

lib/arborist/load-virtual.js
isaacs: hahah, no thanks.  if we start requiring that all code comments be grammatically correct, I'll have a _lot_ of editing to do.

lib/arborist/load-virtual.js
isaacs: `s` is a Shrinkwrap object.  Used consistently that way throughout the file.

lib/arborist/load-virtual.js
d3lm: Haha, no worries 😂 Just something I spotted.

lib/arborist/load-virtual.js
d3lm: Gotcha. Thanks!

PR #120, opened 2020-09-04
fix: install workspaces on update.all=true

ruyadorno:
build-ideal-tree is missing info on workspaces when update.all=true
since the mapping of workspaces got moved to load-actual.

This commit fixes it by reinstating the mapWorkspaces logic to
build-ideal-tree.

fix: https://github.com/npm/cli/issues/1763

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -5,6 +5,7 @@ const pacote = require('pacote')
 const cacache = require('cacache')
 const semver = require('semver')
 const pickManifest = require('npm-pick-manifest')
+const mapWorkspaces = require('@npmcli/map-workspaces')
 const promiseCallLimit = require('promise-call-limit')
 const getPeerSet = require('../peer-set.js')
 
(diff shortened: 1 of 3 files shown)


isaacs:
Once #122 lands, we'll be loading the actual workspaces in loadVirtual and loadActual, so it seems like we only have to do this extra step when it didn't get run for one of those.

Eg, something like this:

```patch
diff --git a/lib/arborist/build-ideal-tree.js b/lib/arborist/build-ideal-tree.js
index 3a8c60f..d0a476c 100644
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -235,13 +235,15 @@ module.exports = cls => class IdealTreeBuilder extends cls {
       // cases we don't use a lockfile anyway.
       // Load on a new Arborist object, so the Nodes aren't the same,
       // or else it'll get super confusing when we change them!
+      // Only have to mapWorkspaces if we didn't get it from actual or virtual
       .then(async root => {
         if (!this[_updateAll] && !this[_global] && !root.meta.loadedFromDisk)
           await new this.constructor(this.options).loadActual({ root })
+        else if (!this[_global] && (!this[_usePackageLock] || this[_updateAll]))
+          await this[_mapWorkspaces](root)
         return root
       })
 
-      .then(tree => this[_mapWorkspaces](tree))
       .then(tree => {
         // null the virtual tree, because we're about to hack away at it
         // if you want another one, load another copy.

```

PR #126, opened 2020-09-11
fix: saveSpec should follow symlinks

ruyadorno:
When saving the ideal tree metadata any installed directory provenant
from a `file:` protocol was using the given filepath to store. This
causes all future calls to `loadActual` to return invalid edges for
loaded links since their fetchSpec won't match what was put in the
stored metadata.

This changes fixes it by making sure we properly follow symlink
realpaths prior to saving the tree metadata.


ruyadorno:
closing this fix in favor of #129 which should be a more proper solution

PR #129, opened 2020-09-12
fix: follow symlinks on build-ideal-tree add

ruyadorno:
Symlinks were being tracked as invalid when loading actual trees since
their realpath wouldn't match, given that build-ideal-tree was just
preserving the fetchSpec read from npm-package-arg when adding new
packages to the ideal tree.

This change fixes it by properly parsing and following symlinks
realpaths at the moment they're added to the ideal tree.


lib/add-rm-pkg-deps.js
```diff
@@ -55,7 +55,7 @@ const addSingle = ({pkg, spec, saveBundle, saveType, path}) => {
   pkg[type] = pkg[type] || {}
   if (rawSpec !== '' || pkg[type][name] === undefined) {
     // if we're in global mode, file specs are based on cwd, not arb path
-    pkg[type][name] = specType === 'file' || specType === 'directory'
+    pkg[type][name] = specType === 'file'
```
ruyadorno, line 58: this was introduced in #125 as a way to compensate for the wrong parsed links and should not be necessary anymore

lib/arborist/build-ideal-tree.js
```diff
@@ -326,18 +337,11 @@ module.exports = cls => class IdealTreeBuilder extends cls {
...
-      const isTag = spec.rawSpec && spec.type === 'tag'
-      return spec.name && !isTag ? spec : pacote.manifest(spec).then(mani => {
-        // if it's a tag type, then we need to run it down to an actual version
-        if (isTag)
-          return npa(`${mani.name}@${mani.version}`)
-
-        spec.name = mani.name
-        return spec
-      })
+      const currpath = this[_global] ? process.cwd() : this.path
```
ruyadorno, line 340: @isaacs I decided to leave this here for now, we can revisit it later

lib/arborist/build-ideal-tree.js
```diff
@@ -355,6 +359,41 @@ module.exports = cls => class IdealTreeBuilder extends cls {
...
+      return spec
+
+    const mani = await pacote.manifest(spec)
+    // if it's a tag type, then we need to run it down to an actual version
+    if (isTag)
+      return npa(`${mani.name}@${mani.version}`)
+
+    spec.name = mani.name
+    return spec
+  }
```
ruyadorno, line 377: This was just a small refactor for readability since when adding the `_followSymlinkPath` method (that is now below) things got gnarly in this part of the code

lib/arborist/build-ideal-tree.js
```diff
@@ -355,6 +359,41 @@ module.exports = cls => class IdealTreeBuilder extends cls {
...
+
+  async [_followSymlinkPath] (spec, currpath) {
+    if(spec.type === 'directory') {
+      const real = await (
+        realpath(spec.fetchSpec, this[_rpcache], this[_stcache])
+          // TODO: create synthetic test case to simulate realpath failure
+          .catch(/* istanbul ignore next */() => null)
```
ruyadorno, line 384: I couldn't get creative enough to come up with a test that would allow us to test for the `realpath` throw without creating a huge mess, I think is fine to leave it as a TODO for later 😬

ruyadorno (commented):
Important thing to note is that I brought this code to my local cli and did a decent round of e2e testing, linking things to global space, linking from there into local projects and overall everything works as expected now 😁

lib/arborist/build-ideal-tree.js
```diff
@@ -355,6 +359,41 @@ module.exports = cls => class IdealTreeBuilder extends cls {
...
+      )
+
+      /* istanbul ignore else - same as the ignored catch above */
+      if (real) {
+        const { name } = spec
+        spec = npa(`file:${relpath(this.path, real)}`, this.path)
```
ruyadorno, line 390: the new spec gets npa parsed relative to `this.path`, so that's what will end up in `resolved`, in the `package.json`, etc - also something I want to validate that makes sense in the impl 😊

lib/add-rm-pkg-deps.js
```diff
@@ -55,7 +55,7 @@ const addSingle = ({pkg, spec, saveBundle, saveType, path}) => {
   pkg[type] = pkg[type] || {}
   if (rawSpec !== '' || pkg[type][name] === undefined) {
     // if we're in global mode, file specs are based on cwd, not arb path
```
isaacs, line 57: Is this comment still valid?  Should be for all file specs, not just global ones.

lib/arborist/build-ideal-tree.js
```diff
@@ -130,6 +135,12 @@ module.exports = cls => class IdealTreeBuilder extends cls {
     this[_linkNodes] = new Set()
     this[_manifests] = new Map()
     this[_peerConflict] = null
+
+    // caches for cached realpath calls
+    const cwd = process.cwd()
+    // assume that the cwd is real enough for our purposes
+    this[_rpcache] = new Map([[ cwd, cwd ]])
```
isaacs, line 142: Since we already have a realpath cache on the Arborist object for use by Arborist.loadActual(), it'd be better for both of them to use `const _rpcache = Symbol.for('realpathCache')`, and only have one of them initialize it.  Otherwise we're doing double work and not caching as efficiently as we could be.

lib/arborist/build-ideal-tree.js
```diff
@@ -130,6 +135,12 @@ module.exports = cls => class IdealTreeBuilder extends cls {
...
+    this[_stcache] = new Map()
```
isaacs, line 143: Same for statCache

lib/arborist/build-ideal-tree.js
isaacs: We can land on a separate commit, but ought to get it fixed, may as well pull in with this changeset.

isaacs (dismissed):
Small nit that we should use single shared `rpcache` and `stcache` members for buildIdealTree and loadActual.

Otherwise LGTM.

lib/arborist/build-ideal-tree.js
ruyadorno: eh, let's land it in a sep commit, maybe even better sep PR, let's thing about future us trying to bisect something 😂

lib/add-rm-pkg-deps.js
ruyadorno: wow this comment is just a lie! will update it

lib/arborist/build-ideal-tree.js
```diff
@@ -326,18 +337,11 @@ module.exports = cls => class IdealTreeBuilder extends cls {
     return Promise.all(add.map(s => {
```
ljharb, line 337: how come this function isn’t `async`, since it returns a Promise? It’d let you use await here.

ruyadorno: I just went with the minimal changes here, I just refactored that first part of the method into `_retrieveSpecName` since it was getting really confusing to follow, but otherwise I'm not biting the bullet to a full `async` refactor, specially in the context of a fix-PR 😊

lib/arborist/build-ideal-tree.js
```diff
@@ -130,6 +135,12 @@ module.exports = cls => class IdealTreeBuilder extends cls {
     this[_linkNodes] = new Set()
     this[_manifests] = new Map()
     this[_peerConflict] = null
+
+    // caches for cached realpath calls
+    const cwd = process.cwd()
+    // assume that the cwd is real enough for our purposes
+    this[_rpcache] = this[_rpcache] || new Map([[ cwd, cwd ]])
+    this[_stcache] = this[_stcache] || new Map()
```
isaacs, line 143: just delete these two lines, no need to re-initialize since ActualLoader does it.

darcyclarke (approved):

Issue #109, opened 2020-08-05
memoize audit metavuln info to disk


isaacs:
Right now, when you have a project with a dependency `p` that has a lot of dependencies, all of which depend back on it, and where it also has a large number of published versions, it can send `npm audit` into a tailspin for quite a long time checking to see which versions of `p` are vulnerable by virtue of their dependence on a vulnerable module `q`.  This is somewhat unavoidable the first time we audit something, but if nothing changes, we have to do all that work _again_ when the user runs `npm audit fix` to fix the problem.

A possible approach to address this would be to cache the vulnerability->metavulnerability relationship somewhere predictable.  The key should be determined such that any change to the advisory version range or available versions of `q` will invalidate it.  Then, when we see a package `p` that depends on `q`, which is vulnerable within `<range>`, we can look in this cached result to see that `p-q-<hash(q versions)>-<hash(range)>` results in vulnerabilities in a set of versions `p`, and a list of the versions of `p` that were available at the time.  Then, we only have to check any additional versions of `p` that we now have, and update the cached data.

This data should probably live in cacache somewhere, since it's not per-project.  That would also help avoid a situation where one project already knows that some version of `p` is vulnerable for some _other_ reason, by seeing any other metavulns applied against it.

isaacs:
https://gist.github.com/isaacs/2060e934476e55deb927bcfe4d444c59

PR #133, opened 2020-09-16
Use @npmcli/metavuln-calculator for faster audits

isaacs:
This makes the AuditReport calculation around 40% faster in the cold
cache state, and anywhere from 90% to 99% faster with a warm cache.

It also sets the stage for the capability of doing audits without
hitting the registry, and potentially avoiding known-vulnerable package
versions if they've been previously marked and cached as vulnerable.
(This additional functionality is not implemented in this change, but
it's something we can do in the future.)

One impact is that the `Vuln` class now has an `advisories` set, as well
as a `via` set.  `vuln.advisories` is a set of advisories created by the
metavuln-calculator, which can represent either actual advisories or
metavulns.  The `via` set is a list of Vuln objects _only_, so that we
can track fixes and report on them comprehensively.

Fix: https://github.com/npm/arborist/issues/109

### DIFF
--- a/lib/audit-report.js
+++ b/lib/audit-report.js
@@ -2,26 +2,13 @@
 
 const npa = require('npm-package-arg')
 const pickManifest = require('npm-pick-manifest')
-const pacote = require('pacote')
-const {intersects, satisfies, simplifyRange} = require('semver')
-const semverOpt = { loose: true, includePrerelease: true }
 
 const Vuln = require('./vuln.js')
+const Calculator = require('@npmcli/metavuln-calculator')
 
 const _getReport = Symbol('getReport')
-const _processDeps = Symbol('processDeps')
-const _processDependent = Symbol('processDependent')
-const _metaVulnSeen = Symbol('metaVulnSeen')
-const _specVulnerableMemo = Symbol('specVulnerableMemo')
-const _addVulnerability = Symbol('addVulnerability')
-const _vulnDependents = Symbol('vulnDependents')
-const _isVulnerable = Symbol('isVulnerable')
-const _specVulnerable = Symbol('specVulnerable')
-const _specVulnCheck = Symbol('specVulnCheck')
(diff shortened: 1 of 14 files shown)


lib/audit-report.js
```diff
@@ -110,112 +100,143 @@ class AuditReport extends Map {
...
-    process.emit('time', 'auditReport:process')
-    for (const p of this[_vulnDependents]) {
-      await this[_processDependent](p)
-    }
-    process.emit('timeEnd', 'auditReport:process')
-  }
+      const vuln = this.get(name) || new Vuln({ name, advisory })
+      if (this.has(name))
+        vuln.addAdvisory(advisory)
+      super.set(name, vuln)
```
ruyadorno, line 145: is there a reason for using `this` (`this.get`, `this.has`) elsewhere but then this one is using `super` (`super.set`) ?

ruyadorno (approved):

lib/audit-report.js
```diff
@@ -110,112 +100,143 @@ class AuditReport extends Map {
...
+
   async [_init] () {
     process.emit('time', 'auditReport:init')
 
     const promises = []
     for (const [name, advisories] of Object.entries(this.report)) {
       for (const advisory of advisories) {
-        const { vulnerable_versions: range } = advisory
-        promises.push(this[_addVulnerability](name, range, advisory))
+        //const { vulnerable_versions: range } = advisory
```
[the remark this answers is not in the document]
isaacs: Hmm, yeah, looks like scar tissue.

lib/audit-report.js
isaacs: `this.set` throws an error, because you're not supposed to use `report.set()` directly.  Calling `super.set()` calls the method from the Map class (which this extends).

ruyadorno: right! I missed the `set()` method! 😄

PR #136, opened 2020-09-17
Add strictPeerDeps, override ERESOLVE if not true

isaacs:
In the overwhelming majority of cases in the wild, a peer dependency
conflict that results in an ERESOLVE can be fixed by using the
`--force` flag.

However, this has other side effects (causing `npm audit fix` to install
semver-major fixes, blowing away file collisions, etc.) which might not
be desirable.  Also, since it's opt-in, it means that users have to run
the install twice for something where we're _pretty_ sure what the right
course of action is.

Let's just make that particular override the default, and reduce most
ERESOLVE errors from a crash to a warning.

(Also, a dumb fix for vuln, should probably pull that in regardless of where we land on this.)

cc: @MylesBorins

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -86,6 +86,7 @@ const _updateFilePath = Symbol('updateFilePath')
 const _followSymlinkPath = Symbol('followSymlinkPath')
 const _getRelpathSpec = Symbol('getRelpathSpec')
 const _retrieveSpecName = Symbol('retrieveSpecName')
+const _strictPeerDeps = Symbol('strictPeerDeps')
 
 // used for the ERESOLVE error to show the last peer conflict encountered
 const _peerConflict = Symbol('peerConflict')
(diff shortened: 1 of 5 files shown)


MylesBorins:
I floated this patch locally and attempted to reify two known failing modules `radium` and `serialport`. Both of those modules now reify without error 🎉

I'm a huge fan of this change, it will give the community some time to adjust and update repos.

ljharb:
@isaacs does this mean that invalid peer deps will no longer fail installs?

ruyadorno (approved):

ruyadorno:
@ljharb it should no longer fail most of the time. It's not the same as `--legacy-peer-deps` though, this change is just making it so that the current behavior we had with `--force` becomes the default, which means: it may still fail in the case of an unsolvable peer dep conflict but the vast majority of cases it will simply resolve the tree assuming the top-most declaration for that peer dep is the one that counts.

It will WARN with [that big pretty-formatted `ERESOLVE` stack](https://github.com/npm/cli/blob/24f3a5448f021ad603046dfb9fd97ed66bd63bba/tap-snapshots/test-lib-utils-explain-eresolve.js-TAP.test.js#L247) explaining precisely what the conflict is, nudging users/lib authors/maintainers (the ecosystem in general) into fixing these borked peer deps down the install tree.

ljharb:
To be clear, when `npm ls` would return nonzero, I would want the install to always fail. Is that property still preserved?

isaacs:
Yes, we'll be setting `strictPeerDeps: true` explicitly in `npm ls`.

ljharb:
@isaacs right but i want `npm install` to fail when `npm ls` would. Is that still the case?

PR #137, opened 2020-09-18
fix: should not count omitted optional deps in diff

ruyadorno:
When using `omit: 'optional'` option the reify diff should not count
optional deps into its diff tree.

Fix: https://github.com/npm/cli/issues/1813


ruyadorno:
I think this also needs to address run-script failures on top of the opt-in `--no-optional` flag, will come back to clean it up

ruyadorno:
will fix it in the cli instead

PR #138, opened 2020-09-18
Don't automatically install optional peer dependencies

bonkydog:
So installing ORMs doesn't install every database driver ever.

## References
https://github.com/npm/rfcs/pull/224


jrylan:
This is awesome, thank you @bonkydog! Also huge thank you to @isaacs for installing peers again!

PR #142, opened 2020-09-24
chore: remove some unused lines on tests

ruyadorno:
Was almost pushing this as part of the other fixes but I feel there's a big chance this is going to be one of these times @isaacs says "oh no, these were actually there for a reason" so a sep PR makes it simpler to just ignore it if that's the case 😄

### DIFF
--- a/test/arborist/build-ideal-tree.js
+++ b/test/arborist/build-ideal-tree.js
@@ -91,8 +91,6 @@ const printTree = tree => ({
   __proto__: { constructor: tree.constructor },
 })
 
-const { format } = require('tcompare')
-
 const cwd = process.cwd()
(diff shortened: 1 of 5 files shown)


isaacs:
LGTM.  Maybe we should lint our tests, too?  I usually don't, because occasionally tests have to do some unsavory things.

ljharb:
Yes, always :-) you can use “overrides” if you need tweaked rules for the tests, as well as override comments.

isaacs:
@ljharb idk, the override we have in place right now is pretty effective.

ljharb:
i took your previous comment as an implication that tests are not linted at all, so you mean, the override that tests aren't linted at all? :-)

PR #146, opened 2020-09-29
fix: workspaces install entering an infinite loop

ruyadorno:
Turns out that having workspaces (or any linked packages) at the same
level depending on conflicting versions of a peer dep was not yet
handled by arborist. This specific combination was setting up the
install into an infinite loop, unable to properly resolve
build-ideal-tree._canPlaceDep.

This changeset adds tests for this specific scenario and also provides a
fix by properly resolving _canPlaceDep to a CONFLICT state if
trying to resolve conflicting versions of a peerDep within a link node,
resulting in throwing a proper ERESOLVE error in these cases.

Fixes: https://github.com/npm/cli/issues/1843

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -1207,10 +1207,13 @@ This is a one-time fix-up, please be patient...
           for (const edge of p.edgesIn) {
             if (edge.peer) {
               // root deps take precedence always.
-              // other peer deps on this node are irrelevant though.
-              if (edge.from.isRoot)
+              // in case this is an edge coming from a link it's also
+              // going to conflict since deps are effectively relative
+              // to its link node parent
+              if (edge.from.isTop)
                 return CONFLICT
 
+              // other peer deps on this node are irrelevant though.
               continue
             }
             // note that we MAY resolve this conflict by using the target's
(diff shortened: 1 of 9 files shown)


lib/arborist/build-ideal-tree.js
```diff
@@ -1206,11 +1206,17 @@ This is a one-time fix-up, please be patient...
           // the target is root, then we can't safely replace.
           for (const edge of p.edgesIn) {
             if (edge.peer) {
-              // root deps take precedence always.
-              // other peer deps on this node are irrelevant though.
-              if (edge.from.isRoot)
+              if (
+                // root deps take precedence always.
```
isaacs, line 1210: Maybe this should just be `isTop` instead of `isRoot`? We probably don't need this check for link targets that aren't top nodes.

isaacs (approved):
One comment to check that might make this a bit cleaner, but if that turns out not to fix the bug, then skip it. Otherwise LGTM 👍

lib/arborist/build-ideal-tree.js
ruyadorno: oh nice top arborist surgeon eye there! it does work all the same 😊 so going with that instead!

PR #148, opened 2020-09-30
Do not run scripts if there is a global top bin conflict

isaacs:
Do not run scripts if there is a global top bin conflict

Fix: https://github.com/github/npm-cli/issues/14

### DIFF
--- a/lib/arborist/rebuild.js
+++ b/lib/arborist/rebuild.js
@@ -22,7 +22,7 @@ const _linkAllBins = Symbol('linkAllBins')
 const _runScripts = Symbol('runScripts')
 const _buildQueues = Symbol('buildQueues')
 const _addToBuildSet = Symbol('addToBuildSet')
-const _checkBins = Symbol('checkBins')
+const _checkBins = Symbol.for('checkBins')
 const _queues = Symbol('queues')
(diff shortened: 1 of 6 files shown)


isaacs:
CI failure due to macOS coveralls push timing out.  Addressed on different branch.

lib/arborist/rebuild.js
```diff
@@ -22,7 +22,7 @@ const _linkAllBins = Symbol('linkAllBins')
 const _runScripts = Symbol('runScripts')
 const _buildQueues = Symbol('buildQueues')
 const _addToBuildSet = Symbol('addToBuildSet')
-const _checkBins = Symbol('checkBins')
+const _checkBins = Symbol.for('checkBins')
```
ljharb, line 25: why use Symbol.for instead of exporting a symbol that can be used explicitly?

darcyclarke (approved):
✅ LGTM

nlf (approved):
looks to get the job done :+1:

lib/arborist/rebuild.js
isaacs: No compelling reason, really.  We use `Symbol.for()` in a few other places as well, when logic has to be shared between mixins.  Would be fine for a later refactor.

PR #149, opened 2020-09-30
Do not drop peer/dev dep while saving if both set

isaacs:
Fix: https://github.com/npm/cli/issues/1849

### DIFF
--- a/test/add-rm-pkg-deps.js
+++ b/test/add-rm-pkg-deps.js
@@ -122,6 +122,56 @@ t.test('add', t => {
     devDependencies: { foo: '2' },
   }, 'update dev')
 
+  t.strictSame(add({
+    pkg: {
+      devDependencies: { foo: '1' },
+      peerDependencies: { foo: '1' },
+    },
+    add: [ foo2 ],
+  }), {
+    devDependencies: { foo: '2' },
+    peerDependencies: { foo: '2' },
+  }, 'update dev and peer together')
+
+  t.strictSame(add({
+    pkg: {
+      devDependencies: { foo: '1' },
+      peerDependencies: { foo: '1' },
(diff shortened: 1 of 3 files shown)


ljharb:
None of the tests have examples of different peer/dep or peer/dev values. What happens there?

isaacs:
@ljharb https://github.com/npm/cli/issues/1849#issuecomment-701561977

The peerDep is the source of truth (since it's loaded first, and we skip the creation of edges if we already have an edge by that name).  When saving, they're both saved to the same value.

isaacs:
Amended in a test showing one with `1` and the other with `*`, both saving to `2`.

test/add-rm-pkg-deps.js
```diff
@@ -122,6 +122,56 @@ t.test('add', t => {
...
+  t.strictSame(add({
+    pkg: {
+      devDependencies: { foo: '1' },
+      peerDependencies: { foo: '*' },
+      peerDependenciesMeta: { foo: { optional: true }},
+    },
+    add: [ foo2 ],
+  }), {
+    devDependencies: { foo: '2' },
+    peerDependencies: { foo: '2' },
```
ljharb, line 171: i would expect a peer of "*" to always remain a peer of "*", otherwise this creates a potentially unintentional breaking change.

isaacs: That's a fair suggestion, but it's not how it works today with `--save-peer`, and changing the status quo there will require some deeper analysis.

darcyclarke (approved):
✅ LGTM

nlf (approved):
i agree with @ljharb that it would make more sense to keep the original peer spec, but given the existing behavior that @isaacs mentioned it seems totally reasonable to land this as-is and revisit that later as a separate issue

PR #150, opened 2020-09-30
Default to building node-gyp dependencies

bonkydog:
npm v6 defaults to building node-gyp dependencies lacking install or preinstall scripts.

It accomplishes this by adding an install script to `package.json`s as they are read.

Here we are moving this logic closer to where we decide to rebuild a dependency.

## References
https://github.com/npm/cli/issues/1848


lib/arborist/rebuild.js
```diff
@@ -178,6 +179,10 @@ module.exports = cls => class Builder extends cls {
       if (bin)
         await this[_checkBins](node)
       set.add(node)
+    } else if (!install && !preinstall && await isNodeGypPackage(node.path)) {
```
isaacs, line 182: Hmm, this skips the `_checkBins` if it's a nodegyp package, and skips the node-gyp default if there's bins.

Could do something like this?

```js
const isGyp = !install && !preinstall && await isNodeGypPackage(node.path)
if (bin || preinstall || install || postinstall || isGyp) {
  if (bin)
    await this[_checkBins](node)
  if (isGyp)
    node.package.scripts.install = 'node-gyp rebuild'
  set.add(node)
}
```

isaacs (changes requested):
Edge case where bins aren't being checked in the "isNodeGyp" case, and node-gyp-ness isn't checked if there are bins.

Other than that, lgtm.

isaacs:
Also, the test fixture being added here is rather large and unwieldy... maybe we could simulate node-report instead of including something that actually creates a binary?

lib/arborist/rebuild.js
bonkydog: Yeah, looks good.  I'll do it!

bonkydog: Done.

bonkydog:
> Also, the test fixture being added here is rather large and unwieldy... maybe we could simulate node-report instead of including something that actually creates a binary?

I swapped node-report out for node-gyp-hello-world. It still does a real build, but it's lighter weight.

PR #151, opened 2020-10-01
Respect gypfile:false flag in package.json

isaacs:
This does a couple of things:

1. Explicitly add a flag in _addToBuildSet to indicate that we've
   already checked this node.  This _should_ be unnecessary, but it's
   possible that we have no script and still need to check for a gypfile
   when we have old metadata, so this addresses the edge case there, and
   makes the logic a little clearer.
2. Use the defaultGypInstallScript from the @npmcli/node-gyp package,
   rather than a hard-coded string.
3. Handle cases where there are bins and also a default node-gyp script.


nlf (approved):
this looks good :+1:

PR #154, opened 2020-10-07
fix: do not rewrite malformed package.json

ruyadorno:
Reify was rewiriting contents of a malformed package.json. This
changeset fixes the problem by making sure build-ideal-tree properly
throws when finding EJSONPARSE errors from the root package.json

Fixes: https://github.com/npm/cli/issues/1883


isaacs (approved):
LGTM!

PR #156, opened 2020-10-12
Isaacs/slightly more strict peer deps

isaacs:
This change makes it so that:

- peerDeps will *always* install when --force is set
- peerDep conflicts will crash when --strict-peer-deps is set
- Otherwise, peerDep conflicts will be overridden force-style whenever
  the *cause* of the conflicting peer deps is not the current root
  project or a workspace.

To accomplish this, the "source" of the peerSet is tracked, and the
ERESOLVE explanation code is changed, because some sorts of conflicts
can be detected in the peerSet creation phase, before a Node is
instantiated.


isaacs:
Landed and pushed in 1.0.0

PR #158, opened 2020-10-14
fix: pass the scope of the package to pacote

nlf:
this allows npm-registry-fetch to correctly identify the registry associated with the request, which resolves the issues with incorrect auth being sent

## References
https://github.com/npm/cli/issues/1959
https://github.com/npm/cli/issues/1960
https://github.com/npm/cli/issues/1800

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -400,7 +400,7 @@ module.exports = cls => class Reifier extends cls {
     // and no 'bundled: true' setting.
     // Do the best with what we have, or else remove it from the tree
     // entirely, since we can't possibly reify it.
-    const res = node.resolved ? this[_registryResolved](node.resolved)
+    const res = node.resolved ? `${node.name}@${this[_registryResolved](node.resolved)}`
       : node.package.name && node.version
         ? `${node.package.name}@${node.version}`
         : null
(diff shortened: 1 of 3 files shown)


isaacs:
🤦

Yes, lgtm.

isaacs:
Just a thought, could we also fix this by changing

```js
const res = node.resolved ? this[_registryResolved](node.resolved)
```

to

```js
const res = node.resolved ? `${node.name}@${this[_registryResolved](node.resolved)}`
```

nlf:
Yup, that also works since it allows npm-package-args to pick up the scope instead

isaacs:
Ok.  Might be safer since it won't clobber a `--scope` config being passed in explicitly?  I'm not sure offhand how that would affect this scenario.

ruyadorno:
I addressed the CR and added a unit test to assert its expectation but I could not really reproduce the initial issues in a complete e2e test in my local machine, therefore I'm not a 100% sure this fixes the bugs as expected. Let me know if you have a repro around and can test it @nlf

nlf:
tested it against my reproduction and it does fix it :+1:

i can't approve a PR i submitted apparently, but it's safe to consider this my approval

PR #159, opened 2020-10-15
fix: workspaces prepare lifecycle scripts

ruyadorno:
When installing workspaces arborist will now run the `prepare` scripts
as part of lifecycle scripts. It also tweaks a bit the build process to
account for placing bin-links AFTER running lifecycle scripts for a
workspace.

Fixes: https://github.com/npm/cli/issues/1965


darcyclarke (approved):
✅ RSLGTM

PR #160, opened 2020-10-15
fix cycles/ordering when peer deps require nesting

isaacs:
Previously, there were cases where a peer dependency group might cycle
back in on itself, in such a way that we added a node and its peers
_prior_ to another member of the peer set that also peer-depended upon
the peer dependency.  Since the nodes were being moved rather than
copied, this meant that the conflicting peer was not around to be
checked, and so we would trigger peer conflict ERESOLVES when we tried
to resolve the dependencies of the second peer dependent at a higher
level.

By copying out of the virtual root peerSet, rather than moving, and
tracking which edges have already been overridden, we can avoid a lot of
other checks later on in the process, while ensuring that the
appropriate set of peer dependencies are being loaded at the optimal
level each time.

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -1023,6 +1023,7 @@ This is a one-time fix-up, please be patient...
       edge: edge.explain(),
       peerConflict,
       strictPeerDeps: this[_strictPeerDeps],
+      force: this[_force],
     }
   }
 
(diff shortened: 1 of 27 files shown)


lib/arborist/build-ideal-tree.js
```diff
@@ -1095,14 +1096,30 @@ This is a one-time fix-up, please be patient...
     // Can only get KEEP here if the original edge was valid,
     // and we're checking for an update but it's already up to date.
     if (canPlace === KEEP) {
-      dep.parent = null
+      if (edge.peer && !target.children.get(edge.name).satisfies(edge)) {
+        // this is an overridden peer dep
+        this[_warnPeerConflict](edge)
+      }
+      // dep.parent = null
```
darcyclarke, line 1103: Any reason we're keeping this around?

darcyclarke (approved):
✅ RSLGTM

PR #161, opened 2020-10-16
Do not over-check peer deps during buildIdealTree 

isaacs:
This caused yargs to fail to install, even when --force was applied!

While building the ideal tree, and checking to see if a node can be
placed at a given location, we also check to ensure that all of that
node's peer dependencies can be placed there.  And, since they can have
peer deps, and those deps can have peer deps, and they may need to be
overridden based on the source dependency's declared intent, we create a
sort of "holding area" in a virtual root.

Previously, in order to check if a peerSet could be placed at a given
location, we just checked all the nodes in this virtual root, and then
moved them into place.  This was a problem, because of the issue
addressed in 94516f3, with cases like those described in
test/fixtures/testing-peer-dep-nesting/README.md in this repository.

In order to address this, the "check all the peers in the set and move
them over" was replaced with a "check all the peers in the set and COPY
them over".  Additionally, since we do still need to be able to dedupe
PART of this peer family up in the tree when possible, the "check all
virtualroot's children" was replaced with an approach where we keep a
list of previously visited members of the peer set, to prevent cycles.

Since there can be multiple peer dependencies on a given dep in the
peer set, some more restrictive than others, so all edges need to be
checked.  (Previously, it would not check meta-peers of the initial node
being placed, so these additional constraints weren't being honored.)

However, since we're re-checking peers of peers within the set, there's
no need to *also* scan through all of the children of the virtualRoot!
Doing so led to an explosion of peer dependency checking, an absolute
cacophony of excessive and damaging pedantic strictness!  As a result,
poor yargs could not install its peer dependencies, even though a
resolution was quite possible without even resorting to force or
non-strict overrides.

Rather than spend several days replicating the precise issue that
caused yargs' dependency graph to trigger a spurious ERESOLVE error,
yargs is added as a test case.  Hence, there are quite a few new
additions to the registry mocks packument data, and many snapshots
changed to reflect the new versions available.  The next commit will
remove all registry mocks which are no longer in use.


ruyadorno (changes requested):
@isaacs I think this might be the time for us to bite the bullet and split the `registry-mocks` into something we can reuse elsewhere?

then we'd be free to add more integration / e2e tests using it too (which I'd love to get in the cli)

ruyadorno (approved):
not going to block the fix on it though 😊 LGTM 👍

PR #162, opened 2020-10-17
fix: symlink on win32 platform without admin rights

salvoravida:


salvoravida:
fix npm7 workspaces symlink on win32

https://github.com/npm/cli/issues/1923

RoystonS:
I think those builds are failing because 100% test coverage isn't being achieved?  That's going to be a hard bar to reach when running tests on Linux and macOS (or any single OS) for a conditional where one side only fires on a different OS (Windows in this case).

salvoravida:
return symlink(rel, node.path, process.platform==="win32" ? 'junction' : 'dir')

i do not expect that any test faild on linux OS, as that change has no effect.

RoystonS:
I don't believe that the test is failing, but that, because the Windows-only path isn't executed, coverage has dropped <100%, so the build is failing because of that.

philipparndt:
I've added some test cases for this in https://github.com/salvoravida/arborist/pull/1
To be able to test this I've extracted a linktype function

```js
linktype (platform = process.platform) {
  return platform === 'win32' ? 'junction' : 'dir'
}
```

salvoravida:
@philipparndt great!

isaacs:
Addressed by #175, no need to check the platform, since the third argument is ignored on posix systems.  Thanks!

PR #179, opened 2020-11-10
Isaacs/update g

isaacs:
Support `npm update -g`
Prevent assignment of `Node.parent` and `Node.fsParent` incorrectly, including in ways that trigger infinite regress when searching for dependency resolutions.


ruyadorno (approved):

jayaddison:
Thanks @isaacs!

PR #183, opened 2020-11-12
Isaacs/test fixture symlink cleanup

isaacs:
based on #182, land that first.


nlf (approved):
:+1: super helpful tweaks

PR #184, opened 2020-11-12
fix: missing bins info in package-lock

ruyadorno:
When migrating from package-lock v1 -> v2 in a project that currently
has an actual tree + a package-lock file the reify process fails to
properly update the package-lock with the bins info.

This fixes it by adding an extra read of `package.json` files of each node of the tree inventory when saving the lockfile v2 after installing from an old lockfile.

Fixes: https://github.com/npm/cli/issues/1957


nlf (approved):
this looks like what we discussed in slack :+1:

PR #187, opened 2020-11-24
fix: allow package.json string shorthands

ruyadorno:
Currently, `bin` and `funding` values from the root `package.json` file are
being normalized when reifying. This changes it to avoid overriding
bin/funding shorthands and always use the original values instead of
their normalized results when writing to the root `package.json` file.

Fix: https://github.com/npm/cli/issues/1964


darcyclarke (approved):
✅ LGTM

PR #189, opened 2020-11-26
fix: support system root path with no pkg name

ruyadorno:
Right now if your cwd is `/` `npm exec`, `npm ls` (or any arborist-dependent
command) will fail with a "could not detect node name from path or
package" error.

This patches add a manual `"system-root"` name to allow arborist to build
a tree from `/`

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -107,13 +107,15 @@ class Node {
 
     this.name = name ||
       nameFromFolder(path || pkg.name || realpath) ||
-      pkg.name
-
-    if (!this.name)
-      throw new TypeError('could not detect node name from path or package')
+      pkg.name ||
+      null
 
     // should be equal if not a link
     this.path = path && resolve(path)
+
+    if (!this.name && (!this.path || this.path !== dirname(this.path)))
+      throw new TypeError('could not detect node name from path or package')
+
     this.realpath = !this.isLink ? this.path : resolve(realpath)
 
     this.resolved = resolved || null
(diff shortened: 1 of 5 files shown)


lib/node.js
```diff
@@ -109,6 +109,10 @@ class Node {
       nameFromFolder(path || pkg.name || realpath) ||
       pkg.name
 
+    // set a manual name if running from fs root dir with no package.json
+    if (!this.name && path === '/')
+      this.name = 'system-root'
```
ljharb, line 114: what happens if i have a dir like `/system-root`, or a package named "system-root"?

ruyadorno: then `this.name` should already be a truthy value (`system-root`) and it would never reach this loc

ljharb: Sure, but is there an issue that two very different paths would result in the same name?

darcyclarke:
### Considerations
- [ ] where can/does this break today?
- [ ] can we fix the references/checks to `this.name` without falling back or using this placeholder?
- [ ] can we use an invalid or scoped package name to ensure there isn't collisions?

lib/node.js
ruyadorno: yeah, it might not be an issue or maybe it is but either way you bring up good points and we decided it's better to wait till next week and have some feedback from @isaacs on it 😊

ruyadorno:
relates to: https://github.com/npm/cli/issues/1975#issuecomment-733928347

isaacs:
Trying a different approach: just let it be nameless if it's in the system root path.  It should be fine, since we never use system root nodes as anything other than a root node, so it's the only case where it ought to be safe to allow a nameless Node instance.

PR #191, opened 2020-12-07
Isaacs/tree check

isaacs:
Pretty massive refactor to the Node class, so that we maintain strict tree integrity at all times.

Adds a `treeCheck` function that will verify tree integrity when in debug mode, so that we can catch issues early and debug them when users run into seemingly impossible scenarios.

Also adds a benchmark script, which can be used to catch performance regressions.  Right now, only `loadActual` and `reify` have benchmark scripts, but it's easy to add more.

I recommend reviewing commit by commit, as the test mocks had to change a bunch, and many deps were updated.


nlf (approved):
this seems good to me!

admittedly i don't have this entire module in my head so my review isn't very nuanced, but i can confirm that tests pass, the code as written makes a reasonable amount of sense, and i'm a big fan of the debug mode tree check method here :+1:

ruyadorno (approved):

PR #195, opened 2020-12-21
fix: optional fsevents dependencies fails on Linux and Windows

Jack-Works:
close https://github.com/npm/cli/issues/2291


isaacs:
https://github.com/npm/cli/issues/2291#issuecomment-754164242

PR #197, opened 2020-12-29
Fixes consistent-resolve.js so that it returns git URLs with the same protocol as originally entered

### LINKED ISSUE #196: [BUG] Consistent resolve.js always returns git+ssh, no matter what protocol was used
# What / Why
An entry in package.json with a `git+https` URL, such as `git+https://git@github.com/a/b.git` is always transformed into an SSH URL by Arborist and saved as `git+ssh://git@github.com/a/b.git`.  This can be a problem if people have specific authentication keys setup for one protocol or another.

## When
* All `git+` URLs get transformed to `git+ssh` instead of retaining their protocol (if specified)

## Where
* URLs specified in package.json when saving to package-lock.json

### Current Behavior
* Add a `git+https` URL to package.json and save the package-lock.json.  The URL will be converted to `git+ssh` in package-lock.json

### Steps to Reproduce
* Add a `git+https` URL to package.json and save the package-lock.json.  The URL will be converted to `git+ssh` in package-lock.json

### Expected Behavior
* A `git+https` URL in package.json is correctly formatted and the protocol is retained in package-lock.json

## References
* Related to https://github.com/npm/cli/issues/2389

drewpc:
Fixes consistent-resolve.js so that it returns git URLs with the same protocol as originally entered

## References
  Fixes #196 and https://github.com/npm/cli/issues/2389 (for NPM 7.x)
  Closes #196

### DIFF
--- a/lib/consistent-resolve.js
+++ b/lib/consistent-resolve.js
@@ -20,7 +20,7 @@ const consistentResolve = (resolved, fromPath, toPath, relPaths = false) => {
     const isPath = type === 'file' || type === 'directory'
     return isPath && !relPaths ? `file:${fetchSpec}`
       : isPath ? 'file:' + (toPath ? relpath(toPath, fetchSpec) : fetchSpec)
-      : hosted ? 'git+' + hosted.sshurl({ noCommittish: false })
+      : hosted ? hosted.toString({ noGitPlus: false, noCommittish: false })
       : type === 'git' ? saveSpec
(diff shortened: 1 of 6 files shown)


isaacs:
Valid issue, but this patch doesn't fix it.  (Explained in https://github.com/npm/cli/issues/2389#issuecomment-754174137)

isaacs:
Fixed by npm/arborist@9989f4f

PR #198, opened 2021-01-04
Add foregroundScripts option

isaacs:
Add foregroundScripts option

This enables us to switch on the old npm v6 behavior of running build
scripts in the foreground, but without the interleaving that occurred in
old npm versions.

When this is enabled, installs will generally go slower and be noisier,
because build scripts inherit stdio and are not run in parallel.
However, for some debugging and logging cases, it is preferable to run
in this mode, despite the added costs.


wesleytodd (approved):

isaacs:
Added the `arb.scriptsRun` set, so we can report on these easily in the CLI.  I think this is ready to land now.

ruyadorno (approved):

PR #199, opened 2021-01-05
fix: should create prefix path if missing

ruyadorno:
npm6 used to mkdir paths along the way if `--prefix` folder was missing, this seems like an unintentional change so this change fixes it by adding a proper path validation/mkdirp at the beggining of reify.

Fixes: https://github.com/npm/cli/issues/2012

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -66,6 +66,7 @@ const _copyIdealToActual = Symbol('copyIdealToActual')
 const _addOmitsToTrashList = Symbol('addOmitsToTrashList')
 const _packageLockOnly = Symbol('packageLockOnly')
 const _dryRun = Symbol('dryRun')
+const _validatePath = Symbol('validatePath')
 const _reifyPackages = Symbol('reifyPackages')
 
 const _omitDev = Symbol('omitDev')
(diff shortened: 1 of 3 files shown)


ljharb:
I'm not sure if this is handled or not, but please make sure the mkdirp only runs if the command would have succeeded - iow, if the command fails prior to fs operations, or is a dry-run, it probably shouldn't alter the filesystem at all?

ruyadorno:
Great points @ljharb! I can def see references to dry-run, package-lock-only and global checks in the other methods so I'll add these checks here too.

Now given that the folder creation is a prerequisite for the other reify methods to work, it's hard for me to see how we could _skip_ mkdirp in these cases, we could maybe rimraf the created folders but then that sounds like possibly more disruptive than just letting the mkdirp behavior be, even though the install failed - remember this mkdirp behavior will **only** apply when users are using a custom `prefix` config pointing to a not existing folder, otherwise folder paths should be there already.

ljharb:
Makes sense - I'm not asking for the folders to never be created unless it succeeds, since that's impractical, but more that I hope to only see them created when it's unsurprising :-)

isaacs:
Doesn't the CLI already mkdirp the prefix config folder if it's missing?  That's where it used to be done, maybe that's where the regression was.  It may be cleaner to only do so when we reify, though, since that's the case where it matters.

PR #201, opened 2021-01-11
remove global guard on validatePath

nlf:
this is maybe not the ideal approach, but it resolves the issue identified [here](https://github.com/npm/cli/issues/2277)

## References

fixes https://github.com/npm/cli/issues/2277


isaacs:
I thought this was fixed by e9665c3, no?

isaacs:
Ohhhhh, I see, it was fixed by e9665c3 but not for global installs.

PR #202, opened 2021-01-12
fix removing global packages when prefix is a symlink

nlf:
currently, when your prefix is a symlink, you cannot remove any global dependencies on any operating system.

i was able to track this down to how the diff library iterates children, essentially because the root is a link and the children getter returns an empty map, we never detect that the global module was installed in the first place, so we fail to identify it needs to be removed.

the tweak here makes it so if a Link class represents the global root, we return the target's children instead of an explicit empty map.

## References

Fixes https://github.com/npm/cli/issues/1869

### DIFF
--- a/lib/diff.js
+++ b/lib/diff.js
@@ -72,6 +72,11 @@ const allChildren = node => {
   if (!node)
     return new Map()
 
+  // if the node is a global root, and also a link, then what we really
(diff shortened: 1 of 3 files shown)


isaacs (approved):
LGTM!

PR #206, opened 2021-01-22
Make printable Node/Link output a bit more useful

isaacs:
This tightens up the Edge representation to what is most useful,
displays Links as 'ArboristLink', and makes the code a bit easier to
work with instead of being one giant inline object.

Edges are now printed as `{ from type name@spec [error]}` for edgesIn,
and `{type name@spec [to] [error]}` for edgesOut, since we only care
about the 'from' value when it's an edge in, and only care about 'to'
when it's an edge out.  Usually, an edge will fit on one line, which is
super handy when debugging trees with lots of dependencies.
(Previously, I'd often console.error() them as
`[from.location,type,spec,to && to.location,error]` for this reason.)

It also separates out the pretty printing stuff into a standalone file, so
it's not taking up as much space in the Node class definition, and a bit
easier to maintain going forward.


lib/printable.js
```diff
@@ -0,0 +1,122 @@
...
+    } }`
+  }
+}
+
+const printableTree = (tree, path = []) => {
+  if (path.includes(tree))
+    return { location: tree.location }
+  path.push(tree)
+  const Cls = tree.isLink ? ArboristLink : ArboristNode
+  return new Cls(tree, path)
```
ljharb, line 119:
```suggestion
  const Cls = tree.isLink ? ArboristLink : ArboristNode
  return new Cls(tree, path.concat(tree))
```
rather than mutating the passed-in array?

ruyadorno (approved):
😍

lib/printable.js
isaacs: Meh.  It's only for debugging, and the path is only used internally to this method.  I think it's fine to just have one, and it means we avoid showing the same node more than once ever (otherwise, you could have two sibling Links pointing to the same target, and it'd print it out twice in its entirety).

PR #208, opened 2021-01-26
Rollback when a fatal signal arrives during reification

isaacs:
Rollback when a fatal signal arrives during reification

This also fixes an issue where the rollback did not properly restore the
state if installing a _new_ package, since there was no move required
when creating the sparse tree in that case, resulting in potentially
half-installed deps being left behind.  Found while combing through
testing every rollback for signal handling.

Fix: https://github.com/npm/cli/issues/2312

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -1,5 +1,6 @@
 // mixin implementing the reify method
 
+const onExit = require('../signal-handling.js')
 const pacote = require('pacote')
 const rpj = require('read-package-json-fast')
 const { updateDepSpec } = require('../dep-spec.js')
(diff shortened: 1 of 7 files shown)


lib/signal-handling.js
```diff
@@ -0,0 +1,68 @@
...
+
+// do all of this in a setup function so that we can call it
+// multiple times for multiple reifies that might be going on.
+// Otherwise, Arborist.reify() is a global action, which is a
+// new constraint we'd be adding with this behavior.
+const setup = fn => {
+  const { process } = module.exports
+
+  const sigListeners = { loaded: false }
+  sigListeners.loaded = false
```
ruyadorno, line 14: looks like a redundant line to me

isaacs: oh, right.  Good catch.

lib/arborist/reify.js
```diff
@@ -153,16 +156,62 @@ module.exports = cls => class Reifier extends cls {
...
+      // only call once.  if signal hits twice, we just terminate
+      removeHandler()
+      reifyTerminated = Object.assign(new Error('process terminated'), {
+        signal,
+      })
+      return false
+    })
+
+    let rollback = _rollbackRetireShallowNodes
+
```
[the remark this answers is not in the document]
isaacs: Well, it would be much _more_ repetitive to check along the way through each of these steps.  This function could be DRY'ed up a bit, but only by making it a bit more convoluted to read.

Yes, I'll add a comment, though 😅

PR #209, opened 2021-01-26
Detect conflicts among nested peerOptional deps

isaacs:
Detect conflicts among nested peerOptional deps

We do not normally install optional peerDependencies, so we were
skipping over them when building up the peer set that is evaluated for
placement within the tree.

However, if an optional peer dep _conflicts_ with something in the tree,
then it will be seen as an invalid edge, and attempt resolution.

This results in a problem in the following scenario:

```
root -> (a, x)
a -> b
b -> PEER(k@1)
x -> y
y -> PEEROPTIONAL(k@2)
```

We place a, b, k@1, and x in the root node_modules without incident.

Then, we evaluate y, and did not add `k@2` to its peerSet, because it
is a peerOptional.  Since it does not have any apparent peers to
conflict, it is _also_ placed at the root level.

Then, we look to the invalid edges coming out of `y`, and find `k@2` is
invalid, because it resolves to `k@1`, and there is no place to put it,
because `y` is already in the root level, where the conflict exists, and
we ERESOLVE.

The solution is to include peerOptional edges when creating the peerSet,
but still omit missing peerOptional deps when looking for invalid edges
needing further resolution.

Fix: https://github.com/npm/cli/issues/2265

(based on #208)

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -1032,19 +1032,20 @@ This is a one-time fix-up, please be patient...
   // deps to override, but throw if no preference can be determined.
   async [_loadPeerSet] (node) {
     const peerEdges = [...node.edgesOut.values()]
-      // we only care about peers here, and don't install peerOptionals
-      .filter(e => e.peer && !e.valid && !e.optional)
+      // we typically only install non-optional peers, but we have to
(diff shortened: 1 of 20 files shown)


wraithgar (approved):

nlf (approved):
this makes sense to me!

PR #211, opened 2021-01-27
fix(install): make sure to reload explicitRequests

wraithgar:
When we are placing our deps, if our edge dep was explicitly requested,
we want to re-place it instead of leaving whatever was there before.

This will mean doing something like `npm install foo` will install the
latest version of `foo` that your tree currently allows, instead of
doing nothing if any version already existed

## References
Closes https://github.com/npm/cli/issues/2243

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -1148,6 +1148,7 @@ This is a one-time fix-up, please be patient...
   [_placeDep] (dep, node, edge, peerEntryEdge = null, peerPath = []) {
     if (edge.to &&
         !edge.error &&
+        !this[_explicitRequests].has(edge.name) &&
         !this[_updateNames].includes(edge.name) &&
         !this[_isVulnerable](edge.to))
       return []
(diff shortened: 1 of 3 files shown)


isaacs:
LGTM once it has a test that fails without it and passes with it.

wraithgar:
Test output if I comment out the fix in `build-ideal-tree.js`

```sh
  not ok must match snapshot
    --- expected                                                                                                         
    +++ actual                                                                                                           
    @@ -12,10 +12,9 @@                                                                                                   
           },                                                                                                            
           "location": "node_modules/lodash",                                                                            
           "name": "lodash",                                                                                             
           "path": "{CWD}/test/arborist/build-ideal-tree-existing-dep-that-needs-to-be-reinstalled/node_modules/lodash", 
    -      "resolved": "https://registry.npmjs.org/lodash/-/lodash-3.10.1.tgz",                                          
    -      "version": "3.10.1",                                                                                          
    +      "version": "3.9.1",                                                                                           
         },                                                                                                              
       },                                                                                                                
       "edgesOut": Map {                                                                                                 
         "lodash" => EdgeOut {                                                                                           
```

isaacs (approved):
LGTM.

PR #212, opened 2021-01-28
More 'root is a link' cleanup

isaacs:
There were a few places in buildIdealTree where we do different behavior
if a node is the root.  However, it should really be if the node is
either the root, *or* the node that a Link root targets.

Add Node.isProjectRoot to mean 'the root node, or the target of a root
link node'.


nlf (approved):
lgtm!

PR #213, opened 2021-01-28
Prevent ERESOLVEs caused by loose root dep specs

isaacs:
Given the following dep graph:

```
root -> a@*, b@2
b -> PEER(a@1)
```

when we load the peerSet for `b@2`, we pull in the parent's
spec for that dependency, `a@*`.  This may result in adding `a@2` to the
tree, which conflicts with b's peer dependency.  However, `a@1` would
have _also_ satisfied the root's dependency and _not_ caused a conflict.

When there is no version that can be found to satisfy both dependencies,
then we use the parent's dependency and warn/crash as we would have
previously.  However, with this commit, we are more aggressively
deduplicating the parent/child dependency collision to avoid ERESOLVEs
that are unnecessary.

This increased deduplication results in a lot of churn in the test
snapshot, because it typically means getting an earlier or more
restrictive version of a shared peer dependency.

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -863,7 +863,7 @@ This is a one-time fix-up, please be patient...
 
   // loads a node from an edge, and then loads its peer deps (and their
   // peer deps, on down the line) into a virtual root parent.
-  [_nodeFromEdge] (edge, parent_) {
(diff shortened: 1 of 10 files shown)


ruyadorno (approved):
We went over this in our pair session and it does LGTM 👍

PR #214, opened 2021-01-29
Properly track dev edges on virtual root nodes

isaacs:
If the virtualRoot node has a sourceReference, and that sourceReference
node would get its devDep edges loaded, then the virtualRoot node should
as well.

This caused a very useless ERESOLVE error, which could not be properly
forced away, when installing in a project that had conflicting peer deps
in its dev tree.


nlf (approved):
this makes sense!

PR #215, opened 2021-01-29
fix global update all so it actually updates things

nlf:
this is intended to make `npm update -g` work correctly

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -233,7 +233,7 @@ module.exports = cls => class Reifier extends cls {
     const actualOpt = this[_global] ? {
       ignoreMissing: true,
       global: true,
-      filter: (node, kid) => !node.isRoot && node !== node.root.target
+      filter: (node, kid) => this[_explicitRequests].size === 0 || !node.isProjectRoot
         ? true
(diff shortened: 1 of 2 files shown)


lib/arborist/reify.js
```diff
@@ -233,7 +234,7 @@ module.exports = cls => class Reifier extends cls {
     const actualOpt = this[_global] ? {
       ignoreMissing: true,
       global: true,
-      filter: (node, kid) => !node.isRoot && node !== node.root.target
+      filter: (node, kid) => this[_updateAll] || (!node.isRoot && node !== node.root.target)
```
isaacs, line 237: If it's global, could we just check for `this[_explicitRequests].size === 0`?

isaacs: Would save having to add another shared symbol.

nlf: yeah that would probably be fine, i went with the shared symbol just because it's more clear what we're actually asserting

isaacs: That second conditional can also be shortened to `!node.isProjectRoot` now.

nlf: :+1: updated that conditional to not use the shared symbol and use isProjectRoot

isaacs (approved):
LGTM

PR #216, opened 2021-01-30
lint tests dry tree printing

isaacs:
Only should be one way to print trees.  Get rid of duplication in the tests, and in the scripts.

Based on #214, land that first.


nlf (approved):
omg yes thank you

PR #217, opened 2021-02-01
Install bins properly when global root is a link

isaacs:
Install bins properly when global root is a link

Also fixes up a few other places where we were checking isRoot and
should have been checking isProjectRoot.

Fix: https://github.com/nodejs/build/issues/2525


nlf (approved):
:+1: confirmed bins go in the right spot with a symlinked root

PR #219, opened 2021-02-02
fix: package.json saving optional deps

ruyadorno:
Reify currently duplicates entries listed as `optionalDependencies` in the
users' `package.json` files. While it's working as expected this is
unexpected to a number of users and it also contradicts our own docs on
it:

    Entries in optionalDependencies will override entries of the same
    name in dependencies, so it's usually best to only put in one place.

This patches this UX problem by adding an extra check that will avoid
adding a dependency to the `package.json` `dependencies` object in case
that package is already listed under `optionalDependencies`.

Fixes: https://github.com/npm/cli/issues/2203
Fixes: https://github.com/npm/cli/issues/1886
Fixes: https://github.com/npm/cli/issues/724

### DIFF
--- a/lib/dep-spec.js
+++ b/lib/dep-spec.js
@@ -22,14 +22,20 @@ const updateDepSpec = (pkg, name, newSpec) => {
   return pkg
 }
 
+const shouldAddToDependencies = (pkg, type, key) =>
+  type !== 'dependencies'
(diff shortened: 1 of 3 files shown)


wraithgar (approved):
Code looks good, test tells me everything I need to know.

ruyadorno:
closing in favor of #221

PR #224, opened 2021-02-10
use the idealtree target path if it exists to make sure relpaths are from the correct root

nlf:
currently, at least in windows, if you attempt to globally install a tarball into a symlinked global prefix the relpath generated here is relative to the original path (i.e. `C:\\Program Files\\nodejs`) rather than the realpath (i.e. `C:\\Users\\username\\AppData\\Roaming\\nvm\\v14.15.5`) which when we try to reify later fails. this change makes sure we use the path that we're going to reify against to generate the relative path, making it valid.

i'll work out a test for this later, just wanted to get the change somewhere reviewable


wraithgar (approved):

ruyadorno (approved):
👍 LGTM

PR #225, opened 2021-02-11
Ensure correct flags on shrinkwrapped module deps

isaacs:
When running `Arborist.loadVirtual()` to read the tree from the point of
a tree that contains a shrinkwrap, we were inappropriately setting all
dep flags to `false`, as we would when reading the virtual tree in a
project root.

This patch updates loadVirtual to ensure that it always sets the proper
flags on nodes within the subtree, if the root supplied is not actually
the project root.

As a result, changes had to be made to the process that inflates
old/ancient lockfiles, and reify()'s handling of garbage data needed to
be updated as well (since the cases being tested previously would now
typically prune the invalid data out as extraneous before getting to the
reify step).  If a node DOES make it through to reify() in the
idealTree, but lacks a version or resolved field, then something is very
wrong, and so we now print a warning to the user asking them to re-try
the install or delete their lockfile (which will typically fix the
problem, since it either is a root dep that's being removed and will be
re-added properly, or a metadep problem that can only be fixed with a
full tree rebuild anyway).


ruyadorno (approved):

nlf (approved):
this looks to resolve the issue in my reproduction case :+1:

Issue #223, opened 2021-02-09
[BUG] peerOptionals can cause ERESOLVE in peerSet, but aren't included in tree


isaacs:
For these dependency graphs:

```
# case a
root -> (x, y@1)
x -> PEEROPTIONAL(z)
z -> PEER(y@2)

# case b
root -> (x) PEEROPTIONAL(y@1)
x -> PEEROPTIONAL(y@2)

# case c
root -> (x) PEEROPTIONAL(y@1)
x -> PEER(z)
z -> PEEROPTIONAL(y@2)
```

The peerOptional dependency is included in the peerSet, raising an ERESOLVE conflict at the peerSet generation stage, even though the peerOptional dependencies will not actually be added to the tree.

In `--force` mode, the peerSet generation error is ignored, since we produce a better warning by waiting until we conflict at the tree generation stage.  Since there is no actual conflict, no warning is generated.

But in default mode, we see the error in the peerSet generation, and raise the error early, assuming that there will be no resolution possible.

We should never have a situation where we crash without `--force`, but do not raise a warning with `--force`.

We should never warn about a dependency that we are not going to actually install.

isaacs:
Examples:

```json
{
  "name": "ng-optional-deps",
  "version": "1.0.0",
  "dependencies": {
    "@angular-devkit/build-angular": "^0.1101.4",
    "@angular/compiler-cli": "11.0"
  }
}
```

```json
{
  "name": "ng-peer-optional-repro",
  "version": "1.0.0",
  "dependencies": {
    "ng-packagr": "11.1.3"
  }
}
```

PR #227, opened 2021-02-11
Do not ERESOLVE on peerOptionals not added to tree

isaacs:
Based on #225, land that first.

For these dependency graphs:

    # case a
    root -> (x, y@1)
    x -> PEEROPTIONAL(z)
    z -> PEER(y@2)

    # case b
    root -> (x) PEEROPTIONAL(y@1)
    x -> PEEROPTIONAL(y@2)

    # case c
    root -> (x) PEEROPTIONAL(y@1)
    x -> PEER(z)
    z -> PEEROPTIONAL(y@2)

The peerOptional dependency is included in the peerSet, which would
raise an ERESOLVE conflict at the peerSet generation stage, even though
the peerOptional dependencies will not actually be added to the tree.

To address this, this commit tracks the nodes which are actually
required in the peerSet generation phase, by virtue of being
non-optionally depended upon by a required node in the peerSet.

If a conflict occurs on a node which is not in the required set during
the peerSet generation phase, we ignore it in much the same way that we
would ignore peerSet errors in metadependencies or when --force is used.

Of course, if the peerOptional dependency is _actually_ required, to
avoid a conflict with an invalid resolution present in the tree already,
and there is no suitable placement for it, then ERESOLVE will still be
raised.

Fix: #223
Fix: https://github.com/npm/cli/issues/2667

### DIFF
--- a/test/arborist/build-ideal-tree.js
+++ b/test/arborist/build-ideal-tree.js
@@ -2480,3 +2480,17 @@ t.test('shrinkwrapped dev/optional deps should not clobber flags', t => {
 
   t.end()
 })
+
+t.test('do not ERESOLVE on peerOptionals that are ignored anyway', t => {
+  // this simulates three cases where a conflict occurs during the peerSet
+  // generation phase, but will not manifest in the tree building phase.
+  const base = resolve(fixtures, 'peer-optional-eresolve')
+  const cases = ['a', 'b', 'c']
+  t.plan(cases.length)
+  for (const c of cases) {
+    t.test(`case ${c}`, async t => {
+      const path = resolve(base, c)
+      t.matchSnapshot(await printIdeal(path))
+    })
+  }
+})
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -827,13 +827,18 @@ This is a one-time fix-up, please be patient...
       // +-- z@1
       // But if x and y are loaded in the same virtual root, then they will
       // be forced to agree on a version of z.
+      const required = edge.type === 'peerOptional' ? new Set()
+        : new Set([edge.from])
+      const parent = edge.peer ? virtualRoot : null
       const dep = vrDep && vrDep.satisfies(edge) ? vrDep
-        : await this[_nodeFromEdge](edge, edge.peer ? virtualRoot : null)
+        : await this[_nodeFromEdge](edge, parent, null, required)
+
       /* istanbul ignore next */
       debug(() => {
         if (!dep)
           throw new Error('no dep??')
       })
+
       tasks.push({edge, dep})
     }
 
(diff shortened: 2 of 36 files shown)


tap-snapshots/test-arborist-build-ideal-tree.js-TAP.test.js
```diff
@@ -12474,6 +12476,394 @@ ArboristNode {
...
+      "integrity": "sha512-GYIjOCQ5BFWl0caylWvjYOLrs0JZq0SkAY0WhNegXPXehob14/DjVD3LxwUWIKzgbhfr345ig46H0jFEmdV66Q==",
+      "inventory": Inventory {},
+      "legacyPeerDeps": false,
+      "linksIn": Set {},
+      "location": "node_modules/@isaacs/testing-peer-optional-conflict-a-x",
+      "name": "@isaacs/testing-peer-optional-conflict-a-x",
+      "optional": false,
+      "path": "{CWD}/test/fixtures/peer-optional-eresolve/a/node_modules/@isaacs/testing-peer-optional-conflict-a-x",
+      "peer": false,
+      "realpath": "{CWD}/test/fixtures/peer-optional-eresolve/a/node_modules/@isaacs/testing-peer-optional-conflict-a-x",
```
ruyadorno, line 12506: heads up, looks like `cleanSnapshot` is failing to clean these up in all windows ci targets

ruyadorno (approved):
👍 LGTM

tap-snapshots/test-arborist-build-ideal-tree.js-TAP.test.js
isaacs: Yeah, I missed something there.  Gotta get that fixed before landing, might be a real issue.

test/arborist/build-ideal-tree.js
```diff
@@ -2392,3 +2406,65 @@ t.test('set the current on ERESOLVE triggered by devDeps', async t => {
...
+t.test('do not ERESOLVE on peerOptionals that are ignored anyway', t => {
+  // this simulates three cases where a conflict occurs during the peerSet
+  // generation phase, but will not manifest in the tree building phase.
+  const base = resolve(fixtures, 'peer-optional-eresolve')
+  const cases = ['a', 'b', 'c']
+  t.plan(cases.length)
+  for (const c of cases) {
+    t.test(`case ${c}`, async t => {
+      const path = resolve(base, c)
+      t.matchSnapshot(await buildIdeal(path))
```
nlf, line 2467:
```suggestion
      t.matchSnapshot(await printIdeal(path))
```

isaacs: ahhhh, right, because the print does the cleanup.  Thanks!

Issue #226, opened 2021-02-11
[BUG] case where ERESOLVE persists when `--force` is used


isaacs:
https://github.com/npm/cli/issues/2504

```json
{
  "devDependencies": {
    "react-scripts": "4.0.1",
    "typescript": "^4.1.3"
  }
}
```

PR #228, opened 2021-02-11
Allow --force to override conflicted peerOptional

isaacs:
Based on #227, land that first.

With a dependency graph like this:

```
root -> (a, b@1)
a -> PEEROPTIONAL(b@2)
```

We do not install the peerOptional dependency by default, so even though
`b@2` is included in the peerSet of `a`, it is not added to the tree.

Then, the `b@1` dependency is added to satisfy root's direct dependency
on it, causing the `a -> b@2` edge to become invalid.

We then try to resolve the `a -> b@2` edge, and find that we cannot
place it anywhere, causing an `ERESOLVE` error.

However, because `b@2` is no longer a part of a peerSet sourced on the
`root` node, we miss the chance to detect that it should be overridden,
resulting in an `ERESOLVE` failure even when `--force` is used.

This commit adds the check for `this[_force]` prior to crashing with
ERESOLVE, so that cases that avoid our earlier heuristics still accept
the invalid resolution when `--force` is in effect.

Fix: #226
Fix: https://github.com/npm/cli/issues/2504

### DIFF
--- a/test/arborist/build-ideal-tree.js
+++ b/test/arborist/build-ideal-tree.js
@@ -2480,3 +2480,17 @@ t.test('shrinkwrapped dev/optional deps should not clobber flags', t => {
 
   t.end()
 })
+
+t.test('do not ERESOLVE on peerOptionals that are ignored anyway', t => {
+  // this simulates three cases where a conflict occurs during the peerSet
+  // generation phase, but will not manifest in the tree building phase.
+  const base = resolve(fixtures, 'peer-optional-eresolve')
+  const cases = ['a', 'b', 'c']
+  t.plan(cases.length)
+  for (const c of cases) {
+    t.test(`case ${c}`, async t => {
+      const path = resolve(base, c)
+      t.matchSnapshot(await printIdeal(path))
+    })
+  }
+})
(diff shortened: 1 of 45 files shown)


test/arborist/build-ideal-tree.js
```diff
@@ -2468,3 +2486,90 @@ t.test('do not ERESOLVE on peerOptionals that are ignored anyway', t => {
...
+        continue
+      t.test(type, async t => {
+        const path = t.testdir({
+          'package.json': JSON.stringify(pj(type)),
+        })
+        t.matchSnapshot(await printIdeal(path, { force: true }), 'use the force')
+        t.rejects(printIdeal(path), { code: 'ERESOLVE' }, 'no force')
+      })
+    }
+  })
```
ruyadorno, line 2572: what about a fourth test in which b is an **optional peer dep** of the root asserting that b is not installed at the end?

ruyadorno (approved):
just an idea for a possible useful test, otherwise LGTM 👍

ruyadorno:
oops, it has conflicts now that you started landing the other PRs 😅

test/arborist/build-ideal-tree.js
isaacs: That is covered by https://github.com/npm/arborist/pull/227

PR #229, opened 2021-02-12
fix: deps of linked deps in package-lock

ruyadorno:
When having optional dependencies of a dependency of
a top-level link dependency, e.g:

    root -> LINK(a)
    a -> (b)
    b -> OPTIONAL(c)

    * c marked as extraneous

Any optional dependency (and all its nested nodes) were marked
as extraneous in package-lock due to `calcDepFlags` missing
properly following the target of Link nodes.

This changeset fixes it by properly following the result of the
`treeverse.visit` method that will have already followed any
link targets at that point.

Fixes: https://github.com/npm/cli/issues/2505

### DIFF
--- a/lib/calc-dep-flags.js
+++ b/lib/calc-dep-flags.js
@@ -11,7 +11,7 @@ const calcDepFlags = (tree, resetRoot = true) => {
     tree,
     visit: node => calcDepFlagsStep(node),
     filter: node => node,
-    getChildren: node => [...node.edgesOut.values()].map(edge => edge.to),
+    getChildren: (node, tree) => [...tree.edgesOut.values()].map(edge => edge.to),
   })
   return ret
 }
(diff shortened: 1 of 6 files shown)


ruyadorno:
@isaacs fixup commit fixes snapshots in a few load-actual and shrinkwrap tests which pruned some `"peer": true` lines.

I assume these now missing lines are due to `calcDepFlags` properly traversing the tree and `unsetFlag(to, 'peer')` in some places it wasn't reaching before. 

Would be good to double check that it all makes sense 👍

lib/calc-dep-flags.js
```diff
@@ -11,7 +11,7 @@ const calcDepFlags = (tree, resetRoot = true) => {
     tree,
     visit: node => calcDepFlagsStep(node),
     filter: node => node,
-    getChildren: node => [...node.edgesOut.values()].map(edge => edge.to),
+    getChildren: (node, tree) => [...tree.edgesOut.values()].map(edge => edge.to),
```
isaacs, line 14: This is the optimally unconfusing way to do it, but it tickles me that the whole diff could've been:

```
- getChildren: node => ....
+ getChildren: (_,node) => ....
```

PR #231, opened 2021-02-16
skip optional deps with mismatched platform

nlf:
this solves for the use case where a package has platform dependent child packages and lists each of them as optionalDependencies such that we will only install the optional dependencies that do not fail platform and engine checks.

this follows the same behavior as npm 6 where `--force` will ignore the checks and install all the optional dependencies.

## References

Fixes https://github.com/npm/cli/issues/2707


lib/arborist/reify.js
```diff
@@ -446,7 +448,20 @@ module.exports = cls => class Reifier extends cls {
     process.emit('time', timer)
     this.addTracker('reify', node.name, node.location)
 
+    const { npmVersion, nodeVersion } = this.options
     const p = Promise.resolve()
+      .then(() => {
```
ljharb, line 453: this function might be a bit cleaner if you use `async [_reifyNode] (node) {`, and then use `await` instead of all the nested `.then`s?

nlf: it might, but to be consistent with the rest of this code base i left it as-is. we should refactor this repo to be more await-y at some point

isaacs (changes requested):
just omit if it's optional.  don't force past it.  users can install as a non-optional dep with `--force` in that case.

Brooooooklyn:
So how could I install dependencies for the other platform now? For example build macOS arm64 electron App on macOS x64 machine?

Related: 

https://github.com/napi-rs/node-rs/issues/376
https://github.com/swc-project/swc/issues/2898

Issue #232, opened 2021-02-17
[BUG] should not write package.json file if nothing is being changed


isaacs:
See https://github.com/npm/cli/issues/2714

We write `package.json`, even when there's no changes to be made.

Untested patch that may fix this:

```diff
diff --git a/lib/update-root-package-json.js b/lib/update-root-package-json.js
index 735ebd1..43e2c40 100644
--- a/lib/update-root-package-json.js
+++ b/lib/update-root-package-json.js
@@ -15,11 +15,18 @@ const depTypes = new Set([
   'peerDependencies',
 ])
 
+const parseJsonSafe = json => {
+  try {
+    return parseJSON(json)
+  } catch (er) {
+    return null
+  }
+}
+
 const updateRootPackageJson = async tree => {
   const filename = resolve(tree.path, 'package.json')
-  const originalContent = await readFile(filename, 'utf8')
-    .then(data => parseJSON(data))
-    .catch(() => null)
+  const originalJson = await readFile(filename, 'utf8')
+  const originalContent = parseJsonSafe(originalJson)
 
   const depsData = orderDeps({
     ...tree.package,
@@ -52,7 +59,8 @@ const updateRootPackageJson = async tree => {
   const content = (JSON.stringify(packageJsonContent, null, format) + '\n')
     .replace(/\n/g, eol)
 
-  return writeFile(filename, content)
+  if (content !== originalJson)
+    return writeFile(filename, content)
 }
 
 module.exports = updateRootPackageJson

```

Should do something similar in lib/shrinkwrap.js, so it doesn't save if nothing is changing.

PR #234, opened 2021-02-17
Do not write package.json if nothing changed

isaacs:
Need a similar fix for Shrinkwrap class.

Fix: https://github.com/npm/arborist/issues/232


nlf (approved):
pretty clear win, good test, this looks great :+1:

PR #235, opened 2021-02-17
Isaacs/prefer peer over prod

isaacs:


nlf (approved):
tests here look good and clearly demonstrate the changes :+1:

Issue #236, opened 2021-02-17
[BUG] another peerOptional ERESOLVE case


isaacs:
Missed a case in https://github.com/npm/arborist/issues/223

```
root -> (x, z@1)
x -> PEEROPTIONAL(y)
y -> PEER(z@2)
```

This is subtly similar to case A

<details>
<summary>case A, hidden to not be confusing</summary>

```
# case a
root -> (x, y@1)
x -> PEEROPTIONAL(z)
z -> PEER(y@2)
```

</details>

But, with the names spelled in such a way that the conflicted dep is not present in the peerSet tree until _after_ the not-required peerOptional meta-peer is added.  (They are sorted alphabetically for consistency.)

This code path wasn't being hit previously, and it's what prevents ng-packagr@11.1.3 from being installed alongside typescript@4.  Ie,

```
x = ng-packagr@11.1.3
y = tsickle
z@1 = typescript@4
z@2 = typescript@~3.9.5
```

# What / Why

> n/a

## When

* n/a

## Where

* n/a

### Current Behavior

* n/a

### Steps to Reproduce

* n/a

### Expected Behavior

* n/a

## Who

* n/a

## References

* n/a

isaacs:
Ok, took a bit more work to get a reproduction case that actually is broken in the same way.  Got to it with this:

```
root -> (x, z@1)
x -> PEEROPTIONAL(y) PEER(z@1) PEER(w@1)
w -> PEER(z@1)
y -> PEER(z@2)
```

PR incoming.

PR #237, opened 2021-02-17
Fix additional peerOptional conflict cases

isaacs:
## case D

```
root -> (x, z@1)
x -> PEEROPTIONAL(y)
y -> PEER(z@2)
```

## case E

```
root -> (x, z@1)
x -> PEEROPTIONAL(y) PEER(z@1)
y -> PEER(z@2)
```

## case F

```
root -> (x, z@1)
x -> PEEROPTIONAL(y) PEER(z@1) PEER(w@1)
w -> PEER(z@1)
y -> PEER(z@2)
```

Case F properly reproduces the failure seen by installing
ng-packagr@11.1.3 alongside typescript@4.

Fix: #236

cc: @kyliau


isaacs:
Note, only Case F is actually broken prior to this commit, but I kept the others for completeness, and to ensure we don't regress in those cases.

PR #241, opened 2021-02-22
Feature: Edges have a "bundled" getter for when they represents a bundled dependency

kumavis:
when analyzing a dep graph its useful to see if an Edge is a `bundledDependencies` edge. this is especially useful when trying to understand the provenance of a package's content.

This PR adds a `bundled` getter property to the `Edge.prototype` which returns true if its a `bundledDependencies` edge or not.

## References
  Related to https://github.com/npm/cli/pull/2750

### DIFF
--- a/lib/edge.js
+++ b/lib/edge.js
@@ -87,14 +87,22 @@ class Edge {
 
   // return the edge data, and an explanation of how that edge came to be here
   [_explain] (seen) {
-    const { error, from } = this
+    const { error, from, bundled } = this
     return {
       type: this.type,
       name: this.name,
       spec: this.spec,
+      ...(bundled ? { bundled } : {}),
       ...(error ? { error } : {}),
       ...(from ? { from: from.explain(null, seen) } : {}),
     }
+  }
+
+  get bundled () {
+    if (!this.from)
+      return false
+    const { package: { bundleDependencies = [] } } = this.from
+    return bundleDependencies.includes(this.name)
   }
 
   get workspace () {
--- a/test/edge.js
+++ b/test/edge.js
@@ -440,3 +440,67 @@ t.match(
   },
   'should return a minimal human-readable representation of the edge obj'
 )
+
+const bundleChild = {
+  name: 'bundle-child',
+  edgesOut: new Map(),
+  edgesIn: new Set(),
+  explain: () => 'bundleChild explanation',
+  package: { name: 'bundle-child', version: '1.2.3' },
+  get version () {
+    return this.package.version
+  },
+  isTop: false,
+  parent: top,
+  resolve (n) {
+    return this.parent.resolve(n)
+  },
+  addEdgeOut (edge) {
+    this.edgesOut.set(edge.name, edge)
+  },
+  addEdgeIn (edge) {
+    this.edgesIn.add(edge)
+  },
+}
+
+const bundleParent = {
+  name: 'bundle-parent',
+  edgesOut: new Map(),
+  edgesIn: new Set(),
+  explain: () => 'bundleParent explanation',
+  package: { name: 'bundle-parent', version: '5.6.7', bundleDependencies: ['bundle-child'] },
+  get version () {
+    return this.package.version
+  },
+  isTop: false,
+  parent: top,
+  resolve (n) {
+    return n === 'bundle-child' ? bundleChild : undefined
+  },
+  addEdgeOut (edge) {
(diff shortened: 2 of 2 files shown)


isaacs:
Since it's only checking that the edge.from is the bundler (and not that the node is in a bundle at all), it would be more efficient to check `edge.from.package.bundleDependencies.includes(edge.name)`. This would also mark the edge as a bundling edge, even if it is not resolved.

isaacs:
This _would_ be really handy to make the code in a few places easier to follow, I like the idea a lot.

kumavis:
yeah sounds good, I was thrown off by this part and wasn't sure what the appropriate check was
https://github.com/npm/arborist/blob/6f99a5e7db8fe21ace364bf7533fe5447149cc25/lib/node.js#L435-L440

any situation where it would be bundled but `this.to.name` would differ from whats in `this.from.package.bundleDependencies`?

lib/edge.js
```diff
@@ -92,11 +92,17 @@ class Edge {
...
+      bundled: this.bundled,
       ...(error ? { error } : {}),
       ...(from ? { from: from.explain(null, seen) } : {}),
     }
   }
 
+  get bundled () {
+    const fromBundleDependencies = this.from && this.from.package && this.from.package.bundleDependencies
+    return (fromBundleDependencies && this.to) ? fromBundleDependencies.includes(this.to.name) : false
+  }
```
ljharb, line 104: is this worth memoizing?

kumavis: its currently only used in `explain`

isaacs: `this.from` will always be set unless the edge is in a detached state.

`this.from.package` will always be set (it'll at least be `{}`).

`this.name` is the name of the edge, which will typically always match the name of the node it resolves to.  (A node name doesn't _always_ match its location in `node_modules`, but only because a node might not _be_ in `node_modules`, but in that case, an edge can't resolve to it.)

So, this bit can be simplified to:

```js
get bundled () {
  if (this.error === 'DETACHED')
    return false
  const { bundleDependencies = [] } = this.from
  return bundleDependencies.includes(this.name)
}
```

lib/edge.js
```diff
@@ -92,11 +92,17 @@ class Edge {
       type: this.type,
       name: this.name,
       spec: this.spec,
+      bundled: this.bundled,
```
isaacs, line 95: This gets quite noisy.

I recommend changing the line above to `const { error, from, bundled } = this`, and then change this line to

```js
  ...(bundled ? { bundled } : {}),
```

so it's filtered out like `from` and `error` when not relevant.

test/edge.js
```diff
@@ -440,3 +458,66 @@ t.match(
...
+  edgesOut: new Map(),
+  edgesIn: new Set(),
+  explain: () => 'bundleChild explanation',
+  package: { name: 'bundle-child', version: '1.2.3' },
+  get version () {
+    return this.package.version
+  },
+  isTop: false,
+  parent: top,
+  getBundler () {
```
isaacs, line 473: I don't think we need `getBundler` in the test if the Edge.bundled getter isn't using it, yes?

lib/edge.js
ljharb: or
```
...(bundled && { bundled }),
```

lib/edge.js
kumavis: my extra defensive code here is partially dealing with the tests which are incomplete non-normalized pojos

lib/edge.js
kumavis: i feel more sympathy for v8 wanting consistent object shape than noise in the snapshots, but can do

lib/edge.js
kumavis: > this.name is the name of the edge, which will typically always match the name of the node it resolves to. 

when would it not match the name of the node

kumavis: is it better to check if its in a detached state which has the side effect of not having a `from`, rather than just checking directly for a from?

lib/edge.js
kumavis: did this

test/edge.js
kumavis: done

lib/edge.js
kumavis: did this, but checked directly for `this.from` instead of the error state

kumavis:
neat. im happy with this

PR #244, opened 2021-02-24
Make tests pass on v10.1

isaacs:
Fixes reliance on `readdir(p, {withFileTypes:true})` that was preventing hidden lockfiles from being loaded.

Also polyfilled all the other things (mostly in tests) that had to be modified to make things pass on node 10.1, and added it to the test matrix.


ljharb (approved):
thanks!

package.json
```diff
@@ -47,6 +47,7 @@
   },
   "scripts": {
     "test": "tap",
+    "test-only": "tap",
```
ljharb, line 50:
```suggestion
    "test": "npm run test-only",
    "test-only": "tap",
```
i usually do this, to ensure that there's only one source of truth for the test command

isaacs: Good call.  You should make it `npm run test-only --` with the `--` in there though, so that `npm test -- --some-tap-arg` works as expected.

isaacs: https://github.com/npm/arborist/pull/244/commits/b6d4e55dac81afd306a58db2984b6a7342c43247

PR #249, opened 2021-03-09
Respect link deps when calculating peerDep sets

isaacs:
Previously, we were not including link targets in the virtual trees
where peer dependency sets are calculated.  Additionally, we were still
using the path `/virtual-root` for the virtual node, even though this is
no longer load-bearing.  (And, as of the other change in this PR to the Node
printable output, no longer necessary or particularly helpful for
debugging.)

As a result, a link dependency from the root node like `file:../../foo`
would get resolved against `/virtual-root`, resulting in `/foo`, which
of course does not match any Node in the virtual tree.  The outcome was
an ERESOLVE error where the `current` Node is shown as having no name
or version (because it is an unsatisfied Link).

The solution is two-part.

First, the path of the virtual tree root now matches the path of the
Node that it is sourced from.

Second, any Link children of the source node have their targets mirrored
in the virtual tree, resulting in them being matched appropriately.

The result is that a Link dependency can now properly satisfy a
peerDependency.  Test shows an example of using a Link to a local Node
as a workaround for a peerSet that otherwise would not be resolveable.

This can of course be abused to get around valid peerDep contracts, but
if they user takes it on themselves to use a local fork of a dependency,
we should respect that in buildIdealTree as we do elsewhere.

Fix: npm/cli#2199

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -883,6 +883,8 @@ This is a one-time fix-up, please be patient...
     // create a virtual root node with the same deps as the node that
     // is requesting this one, so that we can get all the peer deps in
     // a context where they're likely to be resolvable.
+    // Note that the virtual root will also have virtual copies of the
+    // targets of any child Links, so that they resolve appropriately.
     const parent = parent_ || this[_virtualRoot](edge.from)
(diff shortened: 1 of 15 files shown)


ruyadorno (approved):
👍 LGTM

PR #254, opened 2021-03-17
fix: honor explicit prefix

### LINKED ISSUE #253: [BUG] (CLI-related) npm install ignores explicit prefix
# What / Why

I install a package with npm using an explicit prefix, npm ignores the prefix and is using `save-prefix` instead.

The reason appears to be the [`npa` call on line 893 (v2.2.8 snippet)](https://github.com/npm/arborist/blob/v2.2.8/lib/arborist/reify.js#L893). Currently, it's only passing `spec` as the argument. This results in an invalid `req` object as a result of the parse. This `req` object is no longer the expected type `range`, but rather `tag`. When it reaches [lines 908-909](https://github.com/npm/arborist/blob/v2.2.8/lib/arborist/reify.js#L908-L909), `range` is evaluated to `prefixRange`.

I think the fix would be to address the `npa` call on line 893 mentioned above. Below is a possible solution:
```diff
    if (this[_resolvedAdd]) {
      const root = this.idealTree
      const pkg = root.package
      for (const { name } of this[_resolvedAdd]) {
-        const req = npa(root.edgesOut.get(name).spec, root.realpath)
+        const req = npa.resolve(name, root.edgesOut.get(name).spec, root.realpath)
```

## When

I install a package with npm using an explicit prefix.

**Example**

```sh
npm i -D typescript@~3.5.0
```

### Current Behavior

After the installation, `package.json` has `"typescript": "^3.5.3"`.

### Expected Behavior

npm should honor the explicit prefix. `package.json` should have `"typescript": "~3.5.3"`.

## References

* npm/cli#636.
* npm/npm-package-arg#55.

jameschensmith:
## Summary

Honors explicit prefix. npm-package-arg was being used incorrectly in reify when saving the ideal tree. Only the spec was being passed in, which resulted in some unusual results (submitted a ticket to address that in npm-package-arg as well).

Once this was addressed, another issue was spotted, where passing no version on a new package results in `*` being added. Previously, the type resolved to `tag`. This now resolves correctly to `range`. When resolving the variable `range` shortly after, though, existing checks are now truthy. Before, it was short-circuited as falsey due to `type`. Also, node-semver returns false when comparing subsets of `*` other than itself (it's documented in the code). The now falsey resolution makes  spec(`*`) used in a user's `package.json` rather than `prefixRange`.

This request addresses the false semver subset check by checking if `spec` is `*`. `*` should still resolve in a `package.json` if `child.version` is falsy for any reason.

In addition, there was a test which had an unusual alias. This fix brought that abnormality to light. This request corrects that test.

## References

Fixes #253.

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -890,7 +890,7 @@ module.exports = cls => class Reifier extends cls {
       const root = this.idealTree
       const pkg = root.package
       for (const { name } of this[_resolvedAdd]) {
-        const req = npa(root.edgesOut.get(name).spec, root.realpath)
+        const req = npa.resolve(name, root.edgesOut.get(name).spec, root.realpath)
         const {rawSpec, subSpec} = req
 
         const spec = subSpec ? subSpec.rawSpec : rawSpec
--- a/test/arborist/reify.js
+++ b/test/arborist/reify.js
@@ -808,11 +808,12 @@ t.test('saving the ideal tree', t => {
       dependencies: {
         a: 'git+ssh://git@github.com:foo/bar#baz',
         b: '',
-        d: 'd@npm:c@1.x <1.9.9',
+        d: 'npm:c@1.x <1.9.9',
         // XXX: should we remove dependencies that are also workspaces?
         e: 'file:e',
         f: 'git+https://user:pass@github.com/baz/quux#asdf',
         g: '',
+        h: '~1.2.3',
       },
       devDependencies: {
         c: `git+ssh://git@githost.com:a/b/c.git#master`,
@@ -891,6 +892,16 @@ t.test('saving the ideal tree', t => {
         },
         parent: tree,
       })
+      new Node({
+        name: 'h',
+        resolved: 'https://registry.npmjs.org/h/-/h-1.2.3.tgz',
+        pkg: {
+          name: 'h',
+          version: '1.2.3',
+        },
+        parent: tree,
+      })
+
       const target = new Node({
         name: 'e',
         pkg: {
@@ -917,12 +928,11 @@ t.test('saving the ideal tree', t => {
         npa('e'),
         npa('f@git+https://user:pass@github.com/baz/quux#asdf'),
         npa('g'),
+        npa('h@~1.2.3'),
       ]
       // NB: these are all going to be marked as extraneous, because we're
       // skipping the actual buildIdealTree step that flags them properly
-      return a[kSaveIdealTree]({
-        savePrefix: '~',
-      })
+      return a[kSaveIdealTree]({})
     }).then(() => {
       t.matchSnapshot(require(path + '/package-lock.json'), 'lock after save')
       t.strictSame(require(path + '/package.json'), {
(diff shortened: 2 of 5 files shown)


lib/arborist/reify.js
```diff
@@ -905,7 +905,7 @@ module.exports = cls => class Reifier extends cls {
           // would allow versions outside the requested range.  Tags and
           // specific versions save with the save-prefix.
           const isRange = (subSpec || req).type === 'range'
-          const range = !isRange || subset(prefixRange, spec, { loose: true })
+          const range = !isRange || subset(prefixRange, spec, { loose: true }) || spec === '*'
```
jameschensmith, line 908: If this is the route to go, thoughts on switching these checks for speed?

```suggestion
          const range = !isRange || spec === '*' || subset(prefixRange, spec, { loose: true })
```

jameschensmith:
Before merging this PR, I would love for the subset fix to be resolved. I created an issue for it (npm/node-semver#374), and will be submitting a PR shortly.

test/arborist/reify.js
```diff
@@ -808,7 +808,7 @@ t.test('saving the ideal tree', t => {
       dependencies: {
         a: 'git+ssh://git@github.com:foo/bar#baz',
         b: '',
-        d: 'd@npm:c@1.x <1.9.9',
```
isaacs, line 811: Yeah, this looks weird.  Probably shouldn't end up with the name in there...

isaacs (changes requested):
So, I don't think this is quite the right way to go.  I might be wrong, and I agree that there is a bug here, but it'd be much easier to evaluate if we start from a failing test (such as the case you suggested), and then figure out exactly where it's going wrong and how to make that test pass.

I'll dig in today and try to get more insight.  At the very least, saving the name on the right-hand-side of the `dependencies` field is definitely a bug, that shouldn't be happening.

isaacs:
Ah, ok, the `|| spec === '*'` part is unnecessary, but now that npm/node-semver#377 has shipped, I think the other bit looks correct.

This patch seems like it fixes the problem, and is pretty close to what you put here.  Want to take a look? https://gist.github.com/isaacs/7d42bb2672905d3b27e70bc4015ca135

jameschensmith:
> Ah, ok, the `|| spec === '*'` part is unnecessary, but now that [npm/node-semver#377](https://github.com/npm/node-semver/pull/377) has shipped, I think the other bit looks correct.
> 
> This patch seems like it fixes the problem, and is pretty close to what you put here. Want to take a look? [gist.github.com/isaacs/7d42bb2672905d3b27e70bc4015ca135](https://gist.github.com/isaacs/7d42bb2672905d3b27e70bc4015ca135)

Yes! That's what I was planning to do after the patch in node-semver. :grin::clap: If you want me to finish the PR up, let me know. Otherwise, let's get this patched! :raised_hands:

jameschensmith:
Went ahead and addressed the requested changes. Very straightforward :blush: Thanks for your reviews today!

isaacs:
This is shipped now!  Will be in next npm release.  I shuffled around the commit history a bit so that semver update is in one, and the functional updates come together after it.  Thanks!

PR #257, opened 2021-03-25
Isaacs/filtered reification

isaacs:
Adds support for a `workspaces` array of names passed to the Arborist constructor, which will limit reification to only those workspace nodes and their graph of dependencies.

Also, uses the filtered Diff behavior to ensure that global installs only make changes within the specific paths indicated by the set of explicit requests (which should be happening anyway, but which has been a bit of a bug factory in the past).

Open to bike shedding on the option name, it is a bit confusing, since it isn't the map of nodes loaded by `mapWorkspaces`, but I wasn't sure what other way to make the reify call more clear.


isaacs:
Found some bugs exploring the edge cases around this.  Have fixes for them, but don't yet have tests for all the fixes.

nlf (approved):
followed the changes well enough to see what's going on, tests told the rest of the story. i think this makes sense!

PR #258, opened 2021-03-30
Isaacs/reify add to workspaces

isaacs:
Support `arborist.buildIdealTree({ add: [pkg], workspaces: ['list', 'of', 'names'] })` to add dependencies to the workspaces, instead of to the root project.

Also fixes the bug where `npm ls` reports workspace links as extraneous, even though they are clearly not extraneous.

Based on https://github.com/npm/arborist/pull/257, land that first.


lib/arborist/build-ideal-tree.js
```diff
@@ -109,7 +111,7 @@ const _peerSetSource = Symbol.for('peerSetSource')
 
 // used by Reify mixin
 const _force = Symbol.for('force')
-const _explicitRequests = Symbol.for('explicitRequests')
+const _explicitRequests = Symbol('explicitRequests')
```
nlf, line 114: total nit, but it probably makes sense to move this up to the private symbols above since it's no longer "used by Reify mixin"

lib/arborist/reify.js
```diff
@@ -37,6 +38,8 @@ const _retireShallowNodes = Symbol.for('retireShallowNodes')
 const _getBundlesByDepth = Symbol('getBundlesByDepth')
 const _registryResolved = Symbol('registryResolved')
 const _addNodeToTrashList = Symbol('addNodeToTrashList')
+const _workspaces = Symbol.for('workspaces')
```
nlf, line 41: again, nit, but this should be below in the "defined by Ideal mixin"

nlf (approved):
shoulder surfed while this was being done, changes make sense, a couple of super super tiny nits but functionally i think this is what we want :+1:

nlf:
checked on the one test failure, it was a socket timeout in the mock registry server which is.. interesting, but nothing of concern.

wraithgar:
Will this require us to pass in all workspaces if we want all workspaces acted on?  The cli treats an empty array as "all workspaces" currently.

I would say no, and hope that once we are crossing the line here to another module it should be very explicit, but just wanted to make sure we're all on the same page here.

nlf:
i feel like we should require passing in all of the workspaces by name to arborist, let the cli do that resolution since it needs to do it for other commands anyway. better to have that code live in one place imo

lib/arborist/build-ideal-tree.js
isaacs: Ah, good point.

isaacs:
The way you tell Arborist to act on all workspaces is you give it an array with the name of all the workspaces in it.

If you don't provide a `workspaces` array, then it'll add/remove to/from the root package.

PR #259, opened 2021-04-01
Isaacs/links to root causing unreportable eresolve

isaacs:
Fix: https://github.com/npm/cli/issues/2896

This provides the `currentEdge` as a thing that can be reported on in the CLI if somehow the `current` and `peerConflict` nodes aren't found.

It also fixes the edge case triggered by having a dependency of `file:.` in the root node's package.json which then is in conflict with a peerDependency of another root dependency, which is that the Link creates a target with the same path as the `root` node, but which is not the `root` node, and results in some truly spectacular splintering.

Based on #258.


lib/link.js
```diff
@@ -23,13 +23,19 @@ class Link extends Node {
...
+      this.target = this.root
+    else {
+      this.target = new Node({
+        ...options,
+        path: realpath,
+        parent: null,
+        fsParent: null,
+        root: this.root,
+      })
+    }
```
nlf, line 38: for future reviewers, this is the meat of this fix and it 100% makes sense :tada:

nlf (approved):

Issue #135, opened 2020-09-17
add trackers for `audit`


isaacs:
Audits can take a while.  Let's put some tracker spinners in there so that the user has some idea what's going on.

PR #264, opened 2021-04-14
feat: adding spinner for audit process

venkatramanareddy:
## Description
Adding a spinner to indicate user that an audit process is in progress
Few options of spinners (https://github.com/sindresorhus/cli-spinners/blob/main/spinners.json). I didn't want to include cli-spinners package as it would be redundant if the style is fixed. Open for discussion.

## References
For #135

### DIFF
--- a/lib/audit-report.js
+++ b/lib/audit-report.js
@@ -15,6 +15,7 @@ const _omit = Symbol('omit')
 const procLog = require('./proc-log.js')
 
 const fetch = require('npm-registry-fetch')
+const logUpdate = require('log-update')
 
 class AuditReport extends Map {
   static load (tree, opts) {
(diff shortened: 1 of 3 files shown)


isaacs:
Ah, so, this module should not be in the business of setting what the output looks like.

The `lib/tracker.js` mix-in class has some logic for calling the appropriate functions exported by the `npmlog` object.  To be perfectly honest, I still feel like that API has too much tight coupling between Arborist and the other parts of npm, but at least it got us the same behavior that npm v6 had, more or less.

You can look in lib/arborist/build-ideal-tree.js and lib/arborist/reify.js for examples of calling `this.addTracker(name, ...)` and `this.finishTracker(name, ...)` when the operation is finished.  It looks like this isn't showing because we stopped passing `npmlog` to the Arborist instance in the cli's `lib/audit.js` command, and maybe it's just finally time to stop using `npmlog` for this in the first place.

fritzy:
As per @isaacs's response, closing this PR.

PR #266, opened 2021-04-15
fix: correctly handle peer conflicts between peers

nlf:
this corrects the case where a project has dependencies that have peer dependencies with different, yet overlapping required ranges such that we correctly identify the resolved version that matches as many needs as possible.

## References

closes npm/cli#2553
closes npm/cli#2999
closes npm/cli#3083

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -1188,6 +1188,16 @@ This is a one-time fix-up, please be patient...
         }
       }
 
+      // There is something present already, and we're not happy about it
+      // See if the thing we WOULD be happy with is also going to satisfy
+      // the other dependents on the current node.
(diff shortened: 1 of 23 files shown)


nlf:
the two failing runs were transient failures during the tests, socket timeouts. since only one event failed while the other passed for the two that failed, this seems safe to ignore.

princemaple:
Is this in npm 7.11.1? I just updated to this version and `ng update` still getting errors.

princemaple:
thanks for the confirmation @talsi

rarkins:
@talsi is there any new or reopened issue to track this? Perhaps best for you to do it as you have a nice clean reproduction?

talsi:
@rarkins i waiting for https://github.com/npm/cli/issues/2553 to reopen

PR #267, opened 2021-04-16
fix: do not flag nodes that are in a shrinkwrap for deletion

nlf:
since when we are changing or adding a node that contains a shrinkwrap, we expand the tree for that node at reify time, it is safe to not flag these nodes for deletion at all.


lib/diff.js
```diff
@@ -144,11 +145,21 @@ const getChildren = diff => {
   // responsible for installing.
   const actualKids = allChildren(actual)
   const idealKids = allChildren(ideal)
+
+  if (ideal && ideal.hasShrinkwrap && !shrinkwrapInflated.has(ideal)) {
+    // coverage disabled because this is difficult to reach in a synthetic test
```
nlf, line 150: i tried every combination i could think of to try to reach this line, and wasn't able to. @isaacs maybe if you see what i'm missing i can adjust the test cases and remove this

isaacs: Turns out that case is impossible to hit.  I traced it through and replaced the istanbul ignore and the `if` with a comment explaining why we're not being careful there.

PR #268, opened 2021-04-19
Properly handle file:. dep when bundle deps present

isaacs:
If a package had a dependency on `file:.` (ie, to create a symbolic link
to the root project within `node_modules`), as well as a
`bundleDependencies` list, the reifier got into a very strange state and
would fail to resolve its promise, causing npm to exit with a `cb()
never called` error.

The solution is:

- Do not do the extract-and-loadActual pass on dependencies that are
  links.
- Do not create sparse tree folders for dependencies that are links.
  (This wasn't a part of the issue, just a weird inefficiency
  encountered along the way.  We created a dir, only to delete it and
  replace with a symlink later.)
- Later on in the process, when we check to see whether a node has
  already been reified by virtue of having bundleDependencies, instead
  of merely looking at the package.json data, actually keep a list of
  nodes reified ahead of time in this way.
- Lastly, do not run `preinstall`, `install`, and `postinstall` scripts
  for links before linking their bins.  We have to run `prepare` on link
  deps prior to linking their bins, but we should still be linking the
  bins ahead of running the other build scripts, like we do for non-link
  deps.

Based on https://github.com/npm/arborist/pull/265, land that first.

### DIFF
--- a/bin/lib/logging.js
+++ b/bin/lib/logging.js
@@ -20,13 +20,21 @@ const levelMap = new Map(levels.reduce((set, level, index) => {
 }, []))
 
 const { inspect, format } = require('util')
+const colors = process.stderr.isTTY
+const magenta = colors ? msg => `\x1B[35m${msg}\x1B[39m` : m => m
 if (loglevel !== 'silent') {
   process.on('log', (level, ...args) => {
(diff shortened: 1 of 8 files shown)


nlf (approved):
this makes sense! test tells the story of what it's fixing

PR #269, opened 2021-04-21
Prevent excessive bundle deps listing in lockfile

isaacs:


nlf (approved):
nice! this one was bugging me, good find and good fix :+1:

PR #270, opened 2021-04-23
test: package-lock gen missing integrity values

ruyadorno:
When generating a `package-lock.json` file out of an existent and valid actual tree based only in the contents of the `node_modules` folder AND no `package-lock.json` OR hidden lockfile currently available in the project folder, the resulting `package-lock.json` files has no `integrity` or `resolved` values.

### Context

From a UX point of view, `package-lock.json` files are meant to be checked in and tracked by source control systems. It's very unexpected and adds extra friction and unnecessary churn to be excluding `integrity` and `resolved` fields from a resulting `package-lock.json` file when basing a reifycation in an actual tree that contains no lockfile (or hidden lockfile) to cache that data.

### Current test result for quick ref:

Notice that the expected `resolved` and `integrity` values are missing. An extraneous (but probably desirable) `license` field is also added in this case, which is very likely to be a separated problem but it would be nice to make sure to include it in the package-lock-based-reify if that's the expected outcome.

```
 ✖ should result in package-lock with integrity and resolved values

  await reify(path)
  t.matchSnapshot(fs.readFileSync(path + '/package-lock.json', 'utf8'), 'should result in package-lock with integrity and resolved values')
----^
})

  --- expected
  +++ actual
  @@ -4,24 +4,22 @@
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
       "": {
  +      "name": "package-lock-from-actual",
         "version": "1.0.0",
         "dependencies": {
           "abbrev": "^1.1.1"
         }
       },
       "node_modules/abbrev": {
         "version": "1.1.1",
  -      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
  -      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
  +      "license": "ISC"
       }
     },
     "dependencies": {
       "abbrev": {
  -      "version": "1.1.1",
  -      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
  -      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
  +      "version": "1.1.1"
       }
     }
   }

  test: test/arborist/reify.js package-lock contents when generated from actual
    tree with no hidden lockfile or package-lock
  at:
    line: 1503
    column: 5
    file: test/arborist/reify.js
    type: Test
  stack: |
    Test.<anonymous> (test/arborist/reify.js:1503:5)
```

### DIFF
--- a/tap-snapshots/test/arborist/reify.js.test.cjs
+++ b/tap-snapshots/test/arborist/reify.js.test.cjs
@@ -15100,6 +15100,36 @@ ArboristNode {
 }
 `
 
+exports[`test/arborist/reify.js TAP package-lock contents when generated from actual tree with no hidden lockfile or package-lock > should result in package-lock with integrity and resolved values 1`] = `
+{
+  "name": "package-lock-from-actual",
+  "version": "1.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "version": "1.0.0",
+      "dependencies": {
+        "abbrev": "^1.1.1"
+      }
+    },
+    "node_modules/abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
+    }
+  },
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
+    }
+  }
+}
+
+`
+
 exports[`test/arborist/reify.js TAP packageLockOnly can add deps > must match snapshot 1`] = `
 {
   "dependencies": {
--- a/test/arborist/reify.js
+++ b/test/arborist/reify.js
@@ -1497,6 +1497,12 @@ t.test('properly update one module when multiple are present', async t => {
   t.equal(JSON.parse(fs.readFileSync(oncepj, 'utf8')).version, '1.4.0')
 })
 
+t.test('package-lock contents when generated from actual tree with no hidden lockfile or package-lock', async t => {
+  const path = fixture(t, 'package-lock-from-actual')
+  await reify(path)
+  t.matchSnapshot(fs.readFileSync(path + '/package-lock.json', 'utf8'), 'should result in package-lock with integrity and resolved values')
+})
+
 t.test('saving should not replace file: dep with version', async t => {
   // need to run in the path, as if the user typed `npm i file:abbrev`
   const cwd = process.cwd()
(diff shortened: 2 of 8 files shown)


isaacs:
There's really no way that we could possibly get that information, without fetching the packages.  Since there is no guarantee that they came from the default registry (or any registry at all), there's no way to know what the resolved and integrity values _should_ be.

ljharb:
Is the integrity value not persisted inside the npm-installed package.json files in node_modules? (like in one of those underscored properties)

ruyadorno:
> Is the integrity value not persisted inside the npm-installed package.json files in node_modules? (like in one of those underscored properties)

not anymore, in npm7 these values are now stored inside the hidden package-lock file

> There's really no way that we could possibly get that information, without fetching the packages. Since there is no guarantee that they came from the default registry (or any registry at all), there's no way to know what the resolved and integrity values _should_ be.

Generating a `package-lock.json` file with no `integrity/resolved` keys sounds like a terrible middle ground solution for this problem. I would much prefer the result to be at the end of the spectrum, either:

- Do not generate any `package-lock.json` as a result
- Ignore the actual tree and fetch all packages again

ljharb:
@ruyadorno ok, but then the values are still available on disk for a lockfile to be generated from, no?

ruyadorno:
> @ruyadorno ok, but then the values are still available on disk for a lockfile to be generated from, no?

sorry @ljharb it's part of the test scenario but I forgot to add to the description that the hidden lockfile is _also_ missing, happens often if your team has a workflow in which it the `node_modules` are checked in to git but not the hidden lockfile 😉 see **npm/cli**

To be clear, this is an edge case, unlikely to hit most of the user base out there but still one that is affecting us and I would prefer having a better user experience around it.

isaacs:
The resolved and integrity values are not present on disk, anywhere.  Our options, when encountering a lockfile-less package tree, are:

1. **Save what we know, even though it's incomplete.**  This is the current behavior.
2. **Ignore the `node_modules` folder, re-build the package tree and update everything.**  This would make `npm install` be effectively the same as `npm update`, which can be surprising.  Also, it'll occasionally be slow, because we're re-fetching and resolving deps that are ostensibly fine as they are.  And it can fail, if you have a dep in `node_modules` that came from some private registry or something.  (That is, `npm install foo --registry=https://first.com; npm install bar --registry=https://second.com` would never work, because some packages come from one registry, and some come from the other.  Or maybe something was previously installed from just a bare tarball or a git repo, and the name/version doesn't exist in any registry at all!)
3. **Re-fetch and resolve any deps that we find in the tree, to try to figure out what their resolved/integrity values _would_ have been.**  This will be slow and fail for the same use cases that (2) fails, but at least it won't _update_ anything in surprising ways.  What's worse, though, is that they might be lies!  You might do `npm install github:isaacs/fork-of-express`, delete all the resolved/integrity info.  Then a subsequent install will grab the integrity and resolved values of the _real_ `express`, and assume it's the same thing.  That would open the door for supply chain attacks, which we don't want.
4. **Go back to saving resolved and integrity values in the `package.json` files.**  But we still have the problem of packages installed with another package manager, unpacked directly in place, etc.

This is all made worse by the fact that even _with_ a lockfile, you might install from a registry that just doesn't report integrity values, and end up in this state.  And in that case, anything we do to fix it will just keep having to be done every time.

I really don't think this is something we can fix.  If you wanna re-build the dep tree, `npm update` will do that.  We could detect an integrity-less package (other than bundled metadeps, of course), and suggest running `npm update` to build a tree that includes those values, but I think the solutions are all worse than the problem, tbh.

ruyadorno:
We just had a follow up chat and come to the conclusion that a better UX when hitting this edge case is going to be to just log a warning message letting the user know that resolved/integrity values couldn't be found for some packages during the install and potentially suggest running `npm update` as an alternative to clean up/update the package-lock file.

We can revisit some of the alternatives proposed in this thread in the future but for now let's keep the current behavior and add a warning msg.

isaacs:
https://github.com/npm/arborist/issues/275

PR #271, opened 2021-04-27
Prevent and eliminate unnecessary duplicates

isaacs:
When the root package tree bundles its dependencies, previously Arborist
would treat that as a normal bundled dependency (rather than a
transitive dependency bundled _by_ a dependency) and refuse to touch it.

This resulted in a strange situation where npm would stubbornly refuse
to modify the copy of npm-registry-fetch nested underneath its pacote
dependency, even when pacote or nrf were updated explicitly, requiring
the deletion of the node_modules tree in order to resolve properly.

A `canDedupe()` method is added to the Node class, and this is used in
multiple previously (slightly) duplicative code paths where we test to
see if we can dedupe out a nested Node in the tree, thus both preventing
and correcting the unnecessary duplicates that would otherwise have
occurred.

This does introduce the bug/feature/change where now installing a new
package at the top level will cause npm to remove anything nested within
the tree that could be deduped against it, as long as that nested
transitive dependency is not contained in a dep bundle or shrinkwrap.

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -47,6 +47,7 @@ const _flagsSuspect = Symbol.for('flagsSuspect')
 const _workspaces = Symbol.for('workspaces')
 const _prune = Symbol('prune')
 const _preferDedupe = Symbol('preferDedupe')
+const _pruneDedupable = Symbol('pruneDedupable')
 const _legacyBundling = Symbol('legacyBundling')
 const _parseSettings = Symbol('parseSettings')
 const _initTree = Symbol('initTree')
@@ -1342,6 +1343,21 @@ This is a one-time fix-up, please be patient...
         // this is an overridden peer dep
         this[_warnPeerConflict](edge)
       }
+
+      // if we get a KEEP in a update scenario, then we MAY have something
+      // already duplicating this unnecessarily!  For example:
+      // ```
+      // root
+      // +-- x (dep: y@1.x)
+      // |   +-- y@1.0.0
+      // +-- y@1.1.0
+      // ```
+      // Now say we do `reify({update:['y']})`, and the latest version is
+      // 1.1.0, which we already have in the root.  We'll try to place y@1.1.0
+      // first in x, then in the root, ending with KEEP, because we already
+      // have it.  In that case, we ought to REMOVE the nm/x/nm/y node, because
+      // it is an unnecessary duplicate.
+      this[_pruneDedupable](target, true)
       return []
     }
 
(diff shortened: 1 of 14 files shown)


tap-snapshots/test/arborist/load-virtual.js.test.cjs
```diff
@@ -7298,9 +7304,106 @@ ArboristNode {
       "version": "5.5.0",
     },
     "tap" => ArboristNode {
+      "bundleDependencies": Array [
```
nlf, line 7307: are we meant to be listing meta deps as part of the bundleDependencies? i thought the intent was for this to only be the direct deps and everything beneath those is bundled implicitly and not listed here?

tap-snapshots/test/arborist/reify.js.test.cjs
```diff
@@ -695,6 +696,7 @@ ArboristNode {
           "type": "prod",
         },
       },
+      "hasShrinkwrap": true,
```
nlf, line 699: curious what this change is about.. were we missing a path somewhere that was supposed to be populating this before? same for the inShrinkwrap above (and both got added in a bunch of places)

tap-snapshots/test/arborist/load-virtual.js.test.cjs
isaacs: Ah, this test is loading from a npm v6 lockfile.  In that case, we only know whether something is in the bundle or not, not who actually bundles it.  In order to avoid having to trace through the entire dep graph and find a minimal bundle set (which might be inaccurate anyway), we just build the node with an effective virtual `package.json` that bundles everything.

There was a bug where this extra bundle listing was done for _new_ lockfiles, resulting in unnecessary lockfile churn, but that's been fixed.

tap-snapshots/test/arborist/reify.js.test.cjs
isaacs: There's a second commit in the PR that adds the shrinkwrap/bundle info to the pretty-printed nodes.  It was a huge pain to track this issue down without adding that, so I landed it as well.

tap-snapshots/test/arborist/load-virtual.js.test.cjs
nlf: ahhh ok :+1:

tap-snapshots/test/arborist/reify.js.test.cjs
nlf: :+1: makes sense, and definitely helpful for debugging

nlf (approved):
:+1: this makes sense!

PR #272, opened 2021-04-27
support resolvable partial intersecting peerSets

isaacs:
Based on https://github.com/npm/arborist/pull/271, land that first

In cases where a peerSet _partially_ overlaps another, we were
previously treating that as a conflict, because not everything in the
peerSet was being replaced.

So for example,

```
root -> (a)
a -> PEER(w@1, x@1, y@1)
b -> PEER(x@1.0, y@1.0, z@1.0)
```

First we install, and get `w@1.1.0`, `x@1.1.0` and `z@1.1.0`.

Then, we try to install `b`, and need to install the `1.0.0` versions of
`x` and `y`.  We could replace `x` and `y`, but then `z` had no
replacement, and was treated as an ERESOLVE mistakenly.

Some work was required to make it still note when a not-replaced member
of the peer set _is_ indicative of a conflict, because it depends on a
version of a dep within the peer set that is not satisfied by the one
being replaced.

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -1631,32 +1631,137 @@ This is a one-time fix-up, please be patient...
         // placed here as well.  the virtualRoot already has the appropriate
         // overrides applied.
         if (peerEntryEdge) {
-          const peerSet = getPeerSet(current)
-          OUTER: for (const p of peerSet) {
-            // if any have a non-peer dep from the target, or a peer dep if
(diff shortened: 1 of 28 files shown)


nlf (approved):
:+1: looks great!

nlf:
this closes npm/cli#3152

isaacs:
Marking as draft for the moment to work out another edge case this introduces.

nlf (approved):
this turned into a bit of a doozy, but the changes make sense and the comments are _super_ helpful for making this behavior clear

PR #273, opened 2021-04-29
fix(add): properly save new deps where they go

wraithgar:
This consolidates and standardizes the logic around where in the package.json we save added packages.  We explicitly obey the `saveType` if it's given, properly infer it now if it isn't, and enforce exclusivity between things like `dependencies` and `devDependencies`

We also now log a warning when we've ignored one for the other.

## References
Closes: https://github.com/npm/cli/issues/2647
Related to: https://github.com/npm/cli/issues/3096
Related to: https://github.com/npm/cli/issues/2647

PR #276, opened 2021-05-05
Make dependencies' sort locale independent

padinko:
this will fix npm's sorting of dependencies in `package.json` https://github.com/npm/cli/issues/2829

now dependencies in `package.json` are sorted locale dependent, which makes npm unusable on non-english systems

same problem was fixed in 2017 https://github.com/npm/npm/pull/17844/files but now sorting package.json and package-lock.json are again locale independend in npm 7

for me this is a big issue, because some of us in our company is working with english OS, some in slovak and both package.json and package-lock files are changing with every npm install

if you are on *nix system, you can test `LC_ALL=sk npm i` and `LC_ALL=en-US npm i` (as described in issue above). but on windows there is no way to change sort language.

npm 6 from 2017 fix use native `.sort()`, which use `C.UTF-8` sort, now npm 7 use localeCompare, which changed sorting even on English systems: `-` is now after `_`.
so fixing this by forcing sorting to `en-US` locale makes npm 6 and npm 7 dependencies sorted different

### DIFF
--- a/lib/dep-spec.js
+++ b/lib/dep-spec.js
@@ -27,7 +27,7 @@ const orderDeps = (pkg) => {
   for (const type of types) {
     if (pkg && pkg[type]) {
       pkg[type] = Object.keys(pkg[type])
-        .sort((a, b) => a.localeCompare(b))
+        .sort()
         .reduce((res, key) => {
           res[key] = pkg[type][key]
           return res


wraithgar:
At issue here isn't the differences between 6 and 7, its the idempotency of the action.  The results shouldn't differ based on end user locale.

This PR is going to be unable to merge after `https://github.com/npm/arborist/pull/273` is landed, because this function was inlined into `update-root-package-json.js`.

So while I think this is a good change, we'll have to wait for the other PR to land so this one can be re-applied there.  We will also want to have a test that goes inside the `deps are alphabetized` block in `test/update-root-package-json.js` that shows this in action.

isaacs:
Other places where we use `localeCompare`, probably worth moving wholesale to a locale-agnostic sorting throughout npm/cli and its dependencies:

```
lib/config.js
lib/help.js
lib/ls.js
lib/outdated.js
lib/utils/completion/installed-deep.js
lib/utils/config/describe-all.js
lib/utils/npm-usage.js
lib/utils/tar.js
node_modules/@npmcli/arborist/bin/license.js
node_modules/@npmcli/arborist/lib/add-rm-pkg-deps.js
node_modules/@npmcli/arborist/lib/arborist/build-ideal-tree.js
node_modules/@npmcli/arborist/lib/arborist/load-virtual.js
node_modules/@npmcli/arborist/lib/arborist/rebuild.js
node_modules/@npmcli/arborist/lib/audit-report.js
node_modules/@npmcli/arborist/lib/dep-spec.js
node_modules/@npmcli/arborist/lib/printable.js
node_modules/@npmcli/arborist/lib/shrinkwrap.js
node_modules/@npmcli/arborist/lib/vuln.js
node_modules/@npmcli/arborist/lib/yarn-lock.js
node_modules/colors/lib/extendStringPrototype.js
node_modules/glob/common.js
node_modules/ignore-walk/index.js
node_modules/json-stringify-nice/index.js
node_modules/libnpmexec/lib/cache-install-dir.js
node_modules/libnpmexec/lib/index.js
node_modules/npm-packlist/bin/index.js
node_modules/npm-packlist/index.js
scripts/bundle-and-gitignore-deps.js
scripts/config-doc.js
test/lib/exec.js
test/lib/link.js
test/lib/load-all-commands.js
test/lib/utils/cleanup-log-files.js
```

isaacs:
Ok, new plan.

Instead of dropping `localeCompare` for ascii `sort()` order, we're going to specify `'en'` as the second argument to `localeCompare` everywhere.  Internationalization support was added in node 0.12, but was optional until 13, and if not built with full intl support, only the `'en'` locale is used, regardless of the argument.

isaacs:
Going with the approach in #280.  Thanks for bringing this to our attention!

PR #278, opened 2021-05-06
audit: support alias specs and root package names

isaacs:
When the root package has a name like `@magic/semver`, in a folder named
something like `/path/to/magic/semver`, the audit logic would look at
node.name, and see it as `semver`, and then report it as a
vulnerability.

Additionally, a dependency like `npm:mkdirp@0.5.1` would not be detected
as a vulnerability, because the alias spec would never match against the
semver range (assuming that the dependency name even was found as a
vulnerability in the first place).

The fix here is:

1. Add Node.packageName getter, which returns the `name` field from the
  node's package object.
2. Add this field as a queryable field in the inventory.
3. Base audits off of the packageName field, rather than the name field.

Fix: https://github.com/npm/cli/issues/3166


jaeh:
yay :) thanks for all the hard work!

PR #280, opened 2021-05-06
guard against locale-specific sorting

isaacs:
Node will use the current environment's locale sorting for
String.localeCompare.  Thankfully, String.localeCompare can take a
second argument to specify the locale.  Up until Node 13, any valid
argument provided to the function _may_ be respected if Node was
compiled with Intl support, or would always be treated as `'en'`
otherwise.

The solution is to always set the locale to `'en'`.  An alternative
solution would be to compare strings based on byte order, but this does
not provide ideal ergonomis.

To verify this, the locale is set in the test environment to `'sk'`,
which has a significantly different alphabet, including sorting words
starting with `'ch'` _after_ words starting with `'d'`.

Re: npm/cli#2829
Fix: #276

### DIFF
--- a/bin/license.js
+++ b/bin/license.js
@@ -22,7 +22,7 @@ a.loadVirtual().then(tree => {
       set.push([tree.inventory.query('license', license).size, license])
 
     for (const [count, license] of set.sort((a, b) =>
-      a[1] && b[1] ? b[0] - a[0] || a[1].localeCompare(b[1])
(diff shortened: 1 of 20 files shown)


isaacs:
This took a few rounds of updating all the deps in node-tap.  Setting the `LC_ALL=sk` in the test environment shook out a _ton_ of bugs!

wraithgar:
node-tap is updated in `main` so this can be fixed up.

PR #284, opened 2021-05-17
Isaacs/audit workspace

isaacs:
This finishes the "make reify() workspace aware" feature.

1. Dedupe nodes added to/removed from workspaces properly.
2. Make the ideal tree able to add/remove nodes to/from workspaces (not just the project root).
3. Make reify able to properly save added/removed nodes in workspace package.json files.
4. Make audit able to limit the audit report to a specific set of nodes, and build filterSet from workspaces list.


isaacs:
Just changed something that changed a snapshot and caused a line to no longer be covered.  Investigating.

isaacs:
Fixed up the race condition that was resulting in deps being pruned in non-deterministic order.  Ready to review now!

isaacs:
Open question:

Let's say you have a vulnerability in the root project, `json@9.0.0`.  Then you also have a vulnerability in workspace `a`, which has `mkdirp@0.5.0`.  You run `npm audit fix --workspace a`, and I'd expect that only fixes `mkdirp` in the workspace, and leaves `json@9.0.0` in the root project alone.  _But_, reifying also audits.  Should that subsequent audit report on the `json@9.0.0` root dep, since it's still there?  Or no, because you did `--workspace a`?

This could be an issue where you do `npm install -w foo`, and then get audit report containing everything from every workspace.  It's not hard to limit the post-install audit, but it's not how it's working in this PR.

ljharb:
When you do `--workspace a`, *everything* should be scoped to that workspace - including reification and audit warnings. I'd expect it to be mostly the same as if i `cd`d into the directory and ran the same npm command.

isaacs:
> When you do `--workspace a`, imo _everything_ should be scoped to that workspace - including reification and audit warnings. I'd expect it to be mostly the same as if i `cd`d into the directory and ran the same npm command.

Yeah, I agree.  https://github.com/npm/arborist/pull/284/commits/fb2a8b09e39756f082e17a14537322e3c93f5551

ruyadorno (approved):
LGTM 👍

nlf (approved):
this looks good, tests tell a nice story

isaacs:
Found case where Diff would not allow us to prune properly when limited to a workspace.  Fixed on https://github.com/npm/arborist/pull/284/commits/6d71fe76ba00c4552dec84d9410e413977dccb17

Added support to `Arborist.rebuild()` as well.  https://github.com/npm/arborist/pull/284/commits/5aa28d27e101a81e87a4561bbbd2488d7cf41c01

@nlf @ruyadorno If either or both of you could take another look at those last two commits, I'll get this in, and I _think_ this makes Arborist fully workspace-aware.

ruyadorno (approved):

PR #286, opened 2021-05-24
Isaacs/metadata making reify not good

isaacs:
Two commits to make Arborist properly handle some pretty interesting shrinkwrap/metadata handling errors.

### DIFF
--- a/lib/diff.js
+++ b/lib/diff.js
@@ -110,16 +110,33 @@ const getAction = ({actual, ideal}) => {
   if (ideal.isRoot && actual.isRoot)
     return null
 
+  // if the versions don't match, it's a change no matter what
+  if (ideal.version !== actual.version)
+    return 'CHANGE'
+
   const binsExist = ideal.binPaths.every((path) => existsSync(path))
 
   // top nodes, links, and git deps won't have integrity, but do have resolved
-  if (!ideal.integrity && !actual.integrity && ideal.resolved === actual.resolved && binsExist)
+  // However! if *neither* has a resolved value, then we need to refetch
+  // to figure out what it even is, so we mark as CHANGE.
+  const noIntegrity = !ideal.integrity && !actual.integrity
+  const resolvedMatch = ideal.resolved && ideal.resolved === actual.resolved
+  if (noIntegrity && resolvedMatch && binsExist)
+    return null
+
+  // If NEITHER has resolved OR integrity, but the versions match
+  // (which we know because of the check above), and we don't have
+  // to find bins, then we should just leave it as-is.
+  const noResolved = !ideal.resolved && !actual.resolved
+  if (noIntegrity && noResolved && binsExist)
     return null
 
   // otherwise, verify that it's the same bits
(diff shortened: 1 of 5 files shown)


lib/diff.js
```diff
@@ -110,16 +110,33 @@ const getAction = ({actual, ideal}) => {
...
+  // if the versions don't match, it's a change no matter what
+  if (ideal.version !== actual.version)
+    return 'CHANGE'
+
   const binsExist = ideal.binPaths.every((path) => existsSync(path))
 
   // top nodes, links, and git deps won't have integrity, but do have resolved
-  if (!ideal.integrity && !actual.integrity && ideal.resolved === actual.resolved && binsExist)
+  // However! if *neither* has a resolved value, then we need to refetch
+  // to figure out what it even is, so we mark as CHANGE.
```
wraithgar, line 121: This comment seems out of sync w/ the following code block, which does not mark as change.

lib/diff.js
```diff
@@ -110,16 +110,33 @@ const getAction = ({actual, ideal}) => {
...
+  const noIntegrity = !ideal.integrity && !actual.integrity
+  const resolvedMatch = ideal.resolved && ideal.resolved === actual.resolved
+  if (noIntegrity && resolvedMatch && binsExist)
+    return null
+
+  // If NEITHER has resolved OR integrity, but the versions match
+  // (which we know because of the check above), and we don't have
+  // to find bins, then we should just leave it as-is.
+  const noResolved = !ideal.resolved && !actual.resolved
+  if (noIntegrity && noResolved && binsExist)
```
wraithgar, line 131: There are to nearly identical if statements here, that differ only by `resolvedMatch` or `noResolved`.  This may be one of the rare times I advocate for combining the two, and updating the comment to reflect the single conditional.

wraithgar (approved):

lib/diff.js
```diff
@@ -117,18 +117,14 @@ const getAction = ({actual, ideal}) => {
...
-  if (noIntegrity && resolvedMatch && binsExist)
-    return null
-
-  // If NEITHER has resolved OR integrity, but the versions match
-  // (which we know because of the check above), and we don't have
-  // to find bins, then we should just leave it as-is.
   const noResolved = !ideal.resolved && !actual.resolved
-  if (noIntegrity && noResolved && binsExist)
+  const resolvedMatch = ideal.resolved && ideal.resolved === actual.resolved
+  if (noIntegrity && binsExist && (resolvedMatch || noResolved))
```
isaacs, line 127: Wait a minute... I think maybe the comment was saying the right thing, and the code was wrong.

If we don't have any resolved, don't we still need to refetch?  Otherwise we'll never get a resolved value for it?

isaacs: AHH, no, because shrinkwraps might be incomplete.  Verified with the code.  I'll update the comment to prevent us thinking it's a mistake again ;)

wraithgar (approved):

PR #287, opened 2021-06-04
fix: removes [_complete] check to apply correct metadata

gimli01:
removes a check for `[_complete]` which determines whether or not we reify instead unpacking the tarball and looking inside. @isaacs raised some concerns about some slower install times potentially and how this check could be unnecessary. probably want to discuss this as a team.

CC: @darcyclarke 

## References
Closes npm/cli#2606


isaacs:
Need to re-generate snapshots and add a semver tarball to the mock registry.

bl-ue:
>Closes #2606

@gimli01 you mean

>Closes https://github.com/npm/cli/issues/2606

isaacs:
Since we're now _always_ inflating the old (not-ancient) lockfile in buildIdealTree, there's no need to fetch missing package details in reify().  Fixup commit:

- adds missing semver tarball fixture
- updates snapshots
- removes now-dead code in lib/arborist/reify.js

gimli01:
@isaacs looks like one pipeline failed to run

nlf (approved):
this makes sense, the changes all look good. i'm a little curious what the performance impact will be, but since it's likely a one time cost when updating an old lockfile i think that's a really minor concern

nlf:
> @isaacs looks like one pipeline failed to run

one of the tests caused node to crash with exit code `3221225477 ` which translates to `0xC0000005` in hex and represents an access violation. it's pretty likely this failure is due to an issue with the host that ran the test, and not with our code. i went ahead and reran everything though to see if it clears up

PR #290, opened 2021-06-16
fix running unchanged links lifecycle scripts

ruyadorno:
Users expect workspaces (or regular linked deps) to run their prepare
lifecycle script on every reify, since it's often used as some form of
compilation step and should run even though that node hasn't changed.

## References
Fixes: https://github.com/npm/cli/issues/2900

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -1421,7 +1421,7 @@ This is a one-time fix-up, please be patient...
     // prune anything deeper in the tree that can be replaced by this
     if (this.idealTree) {
       for (const node of this.idealTree.inventory.query('name', newDep.name)) {
-        if (node.isDescendantOf(target))
+        if (!node.isTop && node.isDescendantOf(target))
           this[_pruneDedupable](node, false)
       }
     }
@@ -1819,7 +1819,7 @@ This is a one-time fix-up, please be patient...
     const current = target !== entryEdge.from && target.resolve(dep.name)
     if (current) {
       for (const edge of current.edgesIn.values()) {
-        if (edge.from.isDescendantOf(target) && edge.valid) {
+        if (!edge.from.isTop && edge.from.isDescendantOf(target) && edge.valid) {
           if (!edge.satisfiedBy(dep))
             return CONFLICT
         }
@@ -1876,7 +1876,8 @@ This is a one-time fix-up, please be patient...
       if (link.root !== this.idealTree)
         continue
 
-      const external = /^\.\.(\/|$)/.test(relpath(this.path, link.realpath))
+      const tree = this.idealTree.target || this.idealTree
+      const external = !link.target.isDescendantOf(tree)
 
       // outside the root, somebody else's problem, ignore it
       if (external && !this[_follow])
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -914,7 +914,7 @@ This is a one-time fix-up, please be patient...
     await Promise.all(promises)
 
     for (const { to } of node.edgesOut.values()) {
-      if (to && to.isLink)
+      if (to && to.isLink && to.target)
         this[_linkNodes].add(to)
     }
 
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -407,7 +407,7 @@ module.exports = cls => class Reifier extends cls {
     // node.parent is checked to make sure this is a node that's in the tree, and
     // not the parent-less top level nodes
     const filter = node =>
-      node.isDescendantOf(this.idealTree) &&
+      node.top.isProjectRoot &&
         (node.peer && this[_omitPeer] ||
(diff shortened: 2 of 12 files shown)


lib/arborist/reify.js
```diff
@@ -887,6 +887,13 @@ module.exports = cls => class Reifier extends cls {
       filter: diff => diff.action === 'ADD' || diff.action === 'CHANGE',
     })
 
+    // pick up link nodes from the unchanged list as we want to run their
+    // scripts in every install despite of having a diff status change
+    for (const node of this.diff.unchanged) {
+      if (node.isLink)
+        nodes.push(node)
```
isaacs, line 894: It seems like this should be something like `node.isWorkspace` or at least, check that the realpath is within the project root.  Since we don't resolve deps on external link targets, it feels wrong to run their build scripts.

ruyadorno: I didn't wanted to restrict the change to just workspaces since I wanted to make sure we cover the usecase in which users are linking a local module in order to patch something (or just for development purposes) in which they want `prepare` scripts to run

ruyadorno: maybe:

```
      if (node.isLink && node.parent.isTop)
```

instead?

isaacs: Hm, sounds like it's something like:

```js
if (node.isLink) {
  const root = node.root.target || node.root
  if (node.target && node.target.isDescendantOf(root))
    nodes.push(node)
}
```

isaacs: Oh, no, wait, that'll just be `children` not `fsParent`.

Here's how we check for path locality in `buildIdealTree`

```js
      const external = /^\.\.(\/|$)/.test(relpath(this.path, link.realpath))
```

lib/arborist/reify.js
```diff
@@ -887,6 +886,18 @@ module.exports = cls => class Reifier extends cls {
...
 
+    // pick up link nodes from the unchanged list as we want to run their
+    // scripts in every install despite of having a diff status change
+    for (const node of this.diff.unchanged) {
+      const tree = node.root.target || node.root
+
+      // skip links that only live within node_modules as they are most
+      // likely managed by packages we installed, we only want to rebuild
+      // unchanged links we directly manage
+      if (node.isLink && node.target.fsTop === tree)
```
isaacs, line 897: Could also consider `|| !this.diff.unchanged.includes(node.target.fsTop)` to capture the "did not change link, but did change link's fsParent" use case.  (Separate issue from the one this commit addresses, can come later.)

isaacs (approved):
Looks good to me.  Fixes a bug, improves the readability in a few places, and incidentally makes deduping more efficient.  +1, but should maybe get a review from someone who wasn't pairing on it, probably.

lib/arborist/build-ideal-tree.js
```diff
@@ -914,7 +914,7 @@ This is a one-time fix-up, please be patient...
     await Promise.all(promises)
 
     for (const { to } of node.edgesOut.values()) {
-      if (to && to.isLink)
+      if (to && to.isLink && to.target)
```
nlf, line 917: can something be a link and not have a target?

lib/arborist/build-ideal-tree.js
```diff
@@ -1421,7 +1421,7 @@ This is a one-time fix-up, please be patient...
     // prune anything deeper in the tree that can be replaced by this
     if (this.idealTree) {
       for (const node of this.idealTree.inventory.query('name', newDep.name)) {
-        if (node.isDescendantOf(target))
+        if (!node.isTop && node.isDescendantOf(target))
```
nlf, line 1424: this extra check is because it's possible for `isDescendantOf` to return true when the node `isTop`? when is that the case? i'm wondering if it's worth putting a comment here explaining why both tests are needed, because i would sure assume that a node couldn't be both the descendant of a target _and_ a top node

lib/arborist/build-ideal-tree.js
```diff
@@ -1876,7 +1876,8 @@ This is a one-time fix-up, please be patient...
       if (link.root !== this.idealTree)
         continue
 
-      const external = /^\.\.(\/|$)/.test(relpath(this.path, link.realpath))
+      const tree = this.idealTree.target || this.idealTree
+      const external = !link.target.isDescendantOf(tree)
```
nlf, line 1880: this is such a better test than the regex on the relative path, i love it

lib/arborist/reify.js
```diff
@@ -404,10 +404,9 @@ module.exports = cls => class Reifier extends cls {
       return
 
     process.emit('time', 'reify:trashOmits')
-    // node.parent is checked to make sure this is a node that's in the tree, and
-    // not the parent-less top level nodes
+
     const filter = node =>
-      node.isDescendantOf(this.idealTree) &&
+      node.top.isProjectRoot &&
```
nlf, line 409: this is also much more clear to me than the old code

nlf (approved):
a few clarifying questions, but i think this all makes sense

PR #294, opened 2021-06-18
fix(diff): walk target children if root is a link

isaacs:
This expands the behavior that was previously only done for global
roots, but should really be done for _any_ root dir that is not accessed
via its realpath.

Fixes: https://github.com/npm/cli/issues/3431


nlf (approved):
this definitely makes sense, I'm not sure why we had that only for global installs in the first place...

PR #295, opened 2021-06-22
chore: @npmcli/package-json refactor

ruyadorno:
Removes `lib/update-root-package-json.js` in favor of using
`@npmcli/package-json` for reading and modifying `package.json` during
reify.

## References
Relates to: https://github.com/npm/statusboard/issues/368


tap-snapshots/test/arborist/reify.js.test.cjs
```diff
@@ -31952,7 +31952,7 @@ exports[`test/arborist/reify.js TAP save proper lockfile with bins when upgradin
 `
 
 exports[`test/arborist/reify.js TAP save-prod, with optional > must match snapshot 1`] = `
-{"optionalDependencies":{},"dependencies":{"abbrev":"^1.1.1"}}
+{"dependencies":{"abbrev":"^1.1.1"}}
```
wraithgar, line 31955: :chef-kiss:

wraithgar (approved):
Consolidates "write to package json" logic and even fixes an inconsistency in the one-off implementation that was being done here before.

test/update-root-package-json.js
```diff
@@ -1,297 +0,0 @@
...
-      dependencies: undefined,
-      optionalDependencies: {
-        abbrev: '^1.0.0',
-      },
-    },
-    'should write new package.json with tree data'
-  )
-})
-
-t.test('custom formatting', async t => {
```
nlf, line 174: we don't have a test for this in `@npmcli/package-json`, feels like we should add one. i see the code that should do this, but it's always a good idea to prove it works as intended

nlf (approved):
very nice! left one request, but it doesn't belong in this repo and i do not believe needs to hold up this pull request

test/update-root-package-json.js
ruyadorno: we actually have! 😅 I added it as part of https://github.com/npm/package-json/blob/main/test/index.js#L50 which is easy to see in the snapshot: https://github.com/npm/package-json/blob/main/tap-snapshots/test/index.js.test.cjs#L12 but I see how confusing that is from a maintenance point of view! 😬 

I'll swap that test to just use regular formatting and move that check to a `custom formatting` separated test! thanks for bringing that up!

PR #301, opened 2021-07-09
Attempt exposing isDirect and types

naugtur:
This is in relation to  
https://github.com/npm/rfcs/issues/372

Trying to fulfill the first 2 checkboxes from my list here  
https://github.com/npm/rfcs/issues/399

Questions:
Isn't there a nicer way to figure out a direct dependency? I'm not sure if `===''` will work with workspaces, saw examples in tests where the edge was from `package/a` instead of empty string.
Is it ok to add types? They're simply there so I don't have to explore other options outside audit output.

### DIFF
--- a/lib/vuln.js
+++ b/lib/vuln.js
@@ -96,7 +96,7 @@ class Vuln {
   toJSON () {
     const edgesIn = Array.from(this.nodes.values()).flatMap(a => Array.from(a.edgesIn.values())).map(a => a.toJSON())
     const types = new Set(edgesIn.map(a => a.type).sort())
-    const isDirect = edgesIn.any(edge => edge.from === '')
+    const isDirect = [...this.nodes.values()].some(n => [...n.edgesIn].some(e => e.from.isProjectRoot()))
     // Feels like there's a method for checking this and I just didn't find it
 
     return {
--- a/lib/vuln.js
+++ b/lib/vuln.js
@@ -94,10 +94,9 @@ class Vuln {
   }
 
   toJSON () {
-    const edgesIn = Array.from(this.nodes.values()).flatMap(a => Array.from(a.edgesIn.values())).map(a => a.toJSON())
-    const types = new Set(edgesIn.map(a => a.type).sort())
-    const isDirect = [...this.nodes.values()].some(n => [...n.edgesIn].some(e => e.from.isProjectRoot()))
-    // Feels like there's a method for checking this and I just didn't find it
+    const edgesIn = [...this.nodes.values()].flatMap(n => [...n.edgesIn])
+    const isDirect = edgesIn.some(e => e.from.isProjectRoot)
+    const types = [...new Set(edgesIn.map(a => a.type))].sort()
 
     return {
       name: this.name,
--- a/lib/vuln.js
+++ b/lib/vuln.js
@@ -94,10 +94,16 @@ class Vuln {
   }
 
   toJSON () {
-    // sort so that they're always in a consistent order
+    const edgesIn = Array.from(this.nodes.values()).flatMap(a => Array.from(a.edgesIn.values())).map(a => a.toJSON())
+    const types = new Set(edgesIn.map(a => a.type).sort())
+    const isDirect = edgesIn.any(edge => edge.from === '')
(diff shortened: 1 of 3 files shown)


naugtur:
@isaacs @darcyclarke I can see this is fr from ready, but I'm not sure if I'm even going the right direction.

lib/vuln.js
```diff
@@ -94,10 +94,16 @@ class Vuln {
   }
 
   toJSON () {
-    // sort so that they're always in a consistent order
+    const edgesIn = Array.from(this.nodes.values()).flatMap(a => Array.from(a.edgesIn.values())).map(a => a.toJSON())
+    const types = new Set(edgesIn.map(a => a.type).sort())
+    const isDirect = edgesIn.any(edge => edge.from === '')
```
isaacs, line 99:
```suggestion
    const isDirect = [...this.nodes.values()].some(n => [...n.edgesIn].some(e => e.from.isProjectRoot()))
```

naugtur: That's the function I knew must exist but didn't find! Thanks

naugtur:
nyc claims my arrow functions are not covered, which seems like a bug. 
Screenshot 2021-07-11 at 16-21-42 Code coverage report for lib vuln js

Other than that, I think the PR is ready.
Should I add ignore comments or is my understanding of coverage wrong?

lib/vuln.js
```diff
@@ -94,10 +94,15 @@ class Vuln {
   }
 
   toJSON () {
-    // sort so that they're always in a consistent order
+    const edgesIn = [...this.nodes.values()].flatMap(n => [...n.edgesIn])
+    const isDirect = edgesIn.some(e => e.from.isProjectRoot)
+    const types = [...new Set(edgesIn.map(a => a.type))].sort()
```
isaacs, line 99: A node might be "optional", but not have any optional edgesIn.  Or it might have optional edgesIn, but not actually be optional.

For example:

```
root -> (a, b) OPTIONAL(c)
a -> OPTIONAL(d)
b -> (d)
c -> (e)
```

In this dep graph,

- `a` is not an optional dep, but has an optional dep on `d`
- but `b` has a production dep on `d`, so `d` is _not_ optional
- `c` is an optional dep of the root project, and nothing else, so it's optional.
- `e` is a production dep, but a production dep _of_ an optional dep, so it is optional.

Is the intention to show all flags on _any_ nodes in the set, or _all_ nodes in the set?  Eg, if there's one node in the tree that is optional, and another that isn't, should `types` include `'optional'`?

```suggestion
    // this will add the flag to `types` if _any_ dep has the flag set
    const allTypes = ['dev', 'devOptional', 'optional', 'peer']
    const types = new Set()
    // if the semantic should be _all_ deps must have it, do this instead:
    // const types = new Set(allTypes)
    let isDirect = false
    for (const node of this.nodes.values()) {
      for (const t of allTypes) {
        if (node[t])
          types.add(t)
        // if the semantic should be _all_ deps must have the flag set, do this instead:
        // if (!node[t])
        //   types.delete(t)
      }
      // it's a direct dep if the root project or any workspace depends directly on it
      // since workspace projects are "ours" in the same sense as the root project
      // only need to check if we haven't already found one.
      if (!isDirect) {
        for (const { from } of node.edgesIn) {
          if (from.isProjectRoot || from.isWorkspace) {
            isDirect = true
            break
          }
        }
      }
    }
```

isaacs (changes requested):
Main thing is clarifying the semantics around how we set types and isDirect, to ensure that we're getting the correct intended result.

As for the test coverage, the issue is that there are no cases where it's setting anything in `types` or setting `isDirect` to true, so we do need a test case to exercise that logic.

lib/vuln.js
naugtur: I actually only needed dev, so I'm reluctant to include this much complexity here for figuring out types.
I'm going to split this PR so I can finish isDirect and get it merged and use it, then work on types.

naugtur: Extracted here: https://github.com/npm/arborist/pull/323

isaacs:
Closed by #323

PR #303, opened 2021-07-14
feat(rebuild): add lifecycle scripts dependencies

ruyadorno:
Adds the ability for lifecycle scripts of Link nodes to depend
on each other, such that a `prepare` script of linked module **A**
can make use of files that are a result of a `prepare` script of
a linked package **B**.

This is important in order to unlock workflows in which a workspace
needs to make use of another workspace but they all have transpilation
steps (very common in Typescript projects) so tracking the order
(and completion) in which the dependencies lifecycle scripts runs
becomes necessary.

## References
Fixes: https://github.com/npm/cli/issues/3034

### DIFF
--- a/lib/arborist/rebuild.js
+++ b/lib/arborist/rebuild.js
@@ -249,73 +249,162 @@ module.exports = cls => class Builder extends cls {
     if (!queue.length)
       return
 
+    // helper class to manage script execution
+    class Script {
+      constructor (opts) {
+        const {
+          node,
+          timer,
+          ...runScriptOpts
+        } = opts
+        this.node = node
+        this.timer = timer
+        this.opts = runScriptOpts
+        this.pkg = opts.pkg
+        this.path = opts.path
+        this.env = opts.env
+        this.ref = this.node.target.path
+
+        if (!Script.runPromises)
+          Script.runPromises = new Map()
+      }
+
+      // when the script to be run belongs to a link node, this method
+      // retrieves a list of all other link nodes that are direct dependencies
+      // of the current node, their scripts also have to be part of the current
+      // queue, otherwise it just returns an empty list
+      // this is necessary in order to enable workspaces lifecycle scripts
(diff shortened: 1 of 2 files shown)


isaacs (changes requested):
This encounters a deadlock when there is a cycle between link dependencies.  Example to reproduce it:

```patch
diff --git a/test/arborist/rebuild.js b/test/arborist/rebuild.js
index 77c38623..53993201 100644
--- a/test/arborist/rebuild.js
+++ b/test/arborist/rebuild.js
@@ -790,6 +790,7 @@ t.test('scripts dependencies between Link nodes', async t => {
             // at least 20ms prior to writing that file
             prepare: "node -e \"setTimeout(() => { require('fs').writeFileSync(require('path').resolve('index.js'), 'module.exports = function HELLO() {}.name;') }, 20)\"",
           },
+          dependencies: { b: '' },
         }),
       },
       d: {
```

The need is clearly valid, but we have to be able to determine the case where a dependency is waiting on something that exists in a cycle.  Also, I think we can probably address it without needing a class to manage references and polling?  At the very least, the Script class here needs to be a standalone thing with unit tests.

felschr:
We hit this issue afer updating to npm@7.20 and `npm install` suddenly failing because it tries to run `prepare` scripts out of order. I'm happy to see there's already a PR in the pipeline.
With all the other recent additions npm workspaces are starting to become really useful, great work!

alvis:
Maybe what we need is a mechanism to detect a cyclic dependency and raise a warning?

Since 7.20, we can't use workspace with lerna anymore since npm would try to run `prepare` before `leana bootstrap`.
Besides the out-of-order `prepare` sequence, what make it worser is that `npm i --ignore-scripts` doesn't even stop `npm` from running the `prepare` scripts.

fritzy:
@ruyadorno Closing this in order to archive the repo.

PR #305, opened 2021-07-20
Isaacs/refactor build ideal tree

isaacs:
Refactors the `canPlaceDep` and `placeDep` functions out of the `Arborist.buildIdealTree()` mixin class, and into their own classes.

This enables us to track state of all checks more effectively, and resolve certain classes of peerDependency conflicts without the hazard of mutating the tree during the `canPlaceDep` step.  It also makes it possible to track the entire tree of peerDependency set attempts, and for placement attempted, so that we can eventually get to a place where we establish conflict rules and walk back through previous placements pubgrub-style.  (The long term hope there is being able to properly handle cases where `"react-dom": "16||17"` causes `react@17` to be placed, and then a subsequent conflict on `react@16` causes an `ERESOLVE`.)

- [x] clean up commit history
- [ ] docs
- [x] code that works
- [x] tests that cover it fully


lib/shrinkwrap.js
```diff
@@ -183,8 +183,10 @@ const assertNoNewer = async (path, data, lockTime, dir = path, seen = null) => {
         await assertNoNewer(path, data, lockTime, child, seen)
       else if (ent.isSymbolicLink()) {
         const target = resolve(parent, await readlink(child))
-        const tstat = await stat(target).catch(() => null)
+        const tstat = await stat(target).catch(
+          /* istanbul ignore next - windows */ () => null)
```
nlf, line 187: what's the windows issue here? just curious

test/inventory.js
```diff
@@ -1,3 +1,4 @@
+process.env.ARBORIST_DEBUG = '1'
```
nlf, line 1: is this being added to a couple tests deliberate? should it be set in all of the test files, or is this something that we get some specific benefit out of in these ones?

test/can-place-dep.js
```diff
@@ -0,0 +1,1366 @@
...
+    tree,
+    targetLoc,
+    nodeLoc,
+    dep,
+    expect,
+    expectSelf,
+    preferDedupe,
+    peerSet,
+    explicitRequest,
+  }) => {
```
nlf, line 27: i really like how expressive this wrapper makes the tests, but i think it may be useful to put some comments here that explain exactly what each property is for so our future selves don't have to dig too far to remember how to add a new test

nlf (approved):
:tada: just a few clarifying questions/comments, but overall this looks really really great.

_thank you_ for all of the extremely detailed comments, they answered 95% of my questions immediately. this code overall is pretty drastically easier to comprehend, and i'm so happy there's no longer any kind of mutation happening in what should definitely not be a mutating function!

lib/arborist/reify.js
```diff
@@ -642,14 +648,27 @@ module.exports = cls => class Reifier extends cls {
...
+          const arb = new this.constructor({
+            ...this.options,
+            path: node.path,
+          })
+          const notTransplanted = new Set(node.children.keys())
+          await arb.loadActual({
+            root: node,
+            // don't transplant any sparse folders we created
+            transplantFilter: node => {
+              if (node.package._id) {
```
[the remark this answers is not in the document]
isaacs: yeah, the sparse folders we create don't have a package.json, so when we `loadActual` on them, we just get a `node.package` of `{}`.  Anything that has a name and version will also have a `_id`.  We could test for `node.package.name || node.package.version`, but `_id` is that anyway.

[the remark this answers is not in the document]
nlf: :+1: might be worth a comment there explaining

```suggestion
              // loadActual will set node.package to {} for empty directories
              if (node.package._id) {
```

lib/shrinkwrap.js
isaacs: Windows tests fail if you try to create a symlink to a non-existent file and you're not an Admin and haven't set the flag to allow non-Admins to create symbolic links.  Easier to just skip that test on windows, but that means this line won't ever get covered.  Previously, it was getting hit because GH Actions does allow symlinks on windows for non-admins, but I try to keep my windows laptop in a more or less default state, so I found it there.

test/inventory.js
isaacs: I turned off debugging in the buildIdealTree tests on Windows, and everywhere by default on GH Actions, because it's (as of this refactor) just abusively slow to snapshot the entire tree every time we call CanPlaceDep, to the point where tests fail because they time out.

test/can-place-dep.js
isaacs: Good call, will do.

lib/shrinkwrap.js
nlf: ahhhh ok fair enough. so to get coverage there we'd have to have a test that mocks stat with something that rejects and verify that everything went about its business regardless, which being that the thing we're covering is `() => null` that seems pretty excessive

:+1: for just skipping coverage

test/can-place-dep.js
isaacs: https://github.com/npm/arborist/commit/9c53bb310b4a650c8fd9b11413a611b7672ce97c

lib/arborist/reify.js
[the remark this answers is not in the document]
isaacs: 5a3e946cfbb70fd314a0d2ed9f649188a6f0e284

test/inventory.js
isaacs: e9212774d377bbb3ee9c230a8d15b8460f49c128

PR #306, opened 2021-07-31
feat: Save .npmrc files setting workspace-root in workspaces

isaacs:
Related-to: https://github.com/npm/cli/pull/3596

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -12,6 +12,7 @@ const {depth: dfwalk} = require('treeverse')
 const fs = require('fs')
 const {promisify} = require('util')
 const symlink = promisify(fs.symlink)
+const writeFile = promisify(fs.writeFile)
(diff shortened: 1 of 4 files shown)


lib/arborist/reify.js
```diff
@@ -525,6 +530,8 @@ module.exports = cls => class Reifier extends cls {
         }
         await this[_checkBins](node)
         await this[_extractOrLink](node)
+        if (node.isWorkspace && this[_saveWorkspaceRoot])
+          await this[_writeWorkspaceRootConfig](node)
```
ljharb, line 534: this shouldn't be written if the workspace root already exists in the npmrc file - only if it's missing (either the file, or that config value)

fritzy:
@isaacs closing this in order to archive the repo, as we're pulling arborist in as workspace. We're looking at this or something similar.

PR #308, opened 2021-08-10
fix: `npm dedupe` property path of null error

ParadoxInfinite:
Related to https://github.com/npm/cli/pull/3632
Copy pasting from the above closed PR as it's the same change.

> From what I noticed by console logging dep when running npm dedupe, when dep.isLink was true, dep.target was null. Also, dep.realpath exists, so I am not 100% sure, but making an educated guess that we are supposed to use that instead of trying to look for the real path in the target which is null.

Would appreciate feedback, if any!

Also, I would appreciate if someone can help me out with writing a regression test for this? I looked into the `test/place-dep.js` file, but wasn't sure how I could write one for this condition specifically. The way I tested this was using the https://github.com/npm/cli/ repo with the test case provided in the issue below, so any help here is awesome!

## References

Fixes https://github.com/npm/cli/issues/3565

### DIFF
--- a/lib/place-dep.js
+++ b/lib/place-dep.js
@@ -268,7 +268,7 @@ class PlaceDep {
       integrity: dep.integrity,
       legacyPeerDeps: this.legacyPeerDeps,
       error: dep.errors[0],
-      ...(dep.isLink ? { target: dep.target, realpath: dep.target.path } : {}),
+      ...(dep.isLink ? { target: dep.target, realpath: dep.realpath } : {}),
     })
 
     this.oldDep = target.children.get(this.name)


isaacs:
This addresses the surface symptom, and it's a perfectly fine defensive change to make, but doesn't actually make `npm dedupe` behave quite right in the situation presented in npm/cli#3565.  Root caused and fixed in #309 (which incorporates this patch as well).

ParadoxInfinite:
> This addresses the surface symptom, and it's a perfectly fine defensive change to make, but doesn't actually make npm dedupe behave quite right in the situation presented

I understand, as someone who wasn't very familiar with the codebase, this seemed like a logical change as it resolved the issue (or so I thought). I will go through the PR you've mentioned to see if I can understand the issue and the solution better. Hopefully it makes sense to me.

Thank you for taking the time and checking out this PR though!

isaacs:
@ParadoxInfinite You did most of the work of finding it, thanks ;)  Happy to have you keep poking at Arborist, let us know if you have questions.

PR #317, opened 2021-09-14
feat: use Intl.Collator to compare string

killagu:
Refs:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#performance

`Collator#compare` is 100x faster than `String#localeCompare`. And mdn also recommond to use it.

> When comparing large numbers of strings, such as in sorting large arrays, it is better to create an Intl.Collator object and use the function provided by its compare property.

It's the benchmark code and the result.

```
String#localeCompare x 1.50 ops/sec ±2.78% (8 runs sampled)
Collator#compare x 165 ops/sec ±1.00% (83 runs sampled)
Fastest is Collator#compare
```

```js
'use strict';

const Benchmark = require('benchmark');
const cryprto = require('crypto');
const suite = new Benchmark.Suite();

const strCount = 1e5;
const strList1 = [];
const strList2 = [];
for (let i = 0; i < strCount; i++) {
  strList1.push(cryprto.randomBytes(5).toString('hex'));
  strList2.push(cryprto.randomBytes(5).toString('hex'));
}

function useStr() {
  for (let i = 0; i < strCount; i++) {
    strList1[i].localeCompare(strList2[i], 'en');
  }
}

const collator = Intl.Collator('en');
function useIntr() {
  for (let i = 0; i < strCount; i++) {
    collator.compare(strList1[i], strList2[i]);
  }
}

// add tests
suite.add('String#localeCompare', function() {
  useStr();
})
  .add('Collator#compare', function() {
    useIntr();
  })
  // add listeners
  .on('cycle', function(event) {
    console.log(String(event.target));
  })
  .on('complete', function() {
    console.log('Fastest is ' + this.filter('fastest')
      .map('name'));
  })
  // run async
  .run({ 'async': true });
```

### DIFF
--- a/lib/place-dep.js
+++ b/lib/place-dep.js
@@ -19,6 +19,7 @@ const debug = require('./debug.js')
 const Link = require('./link.js')
 const gatherDepSet = require('./gather-dep-set.js')
 const peerEntrySets = require('./peer-entry-sets.js')
+const collator = new Intl.Collator('en');
 
 class PlaceDep {
   constructor (options) {
@@ -430,7 +431,7 @@ class PlaceDep {
       // sort these so that they're deterministically ordered
       // otherwise, resulting tree shape is dependent on the order
       // in which they happened to be resolved.
-      const nodeSort = (a, b) => a.location.localeCompare(b.location, 'en')
+      const nodeSort = (a, b) => collator.compare(a.location, b.location)
 
       const children = [...node.children.values()].sort(nodeSort)
       for (const child of children) {


ljharb:
Which node version came with Intl support by default? npm supports node 10.

Also, what about node compilations that omit Intl?

killagu:
Node 12 start came with Intl. Maybe we could check version to chose string or intl

ljharb:
If there’s a perf benefit, that makes sense - in general, since Intl isn’t in the actual language, code can’t rely on it being present.

isaacs:
If we're going to do this, we should do it all the way, and with caching and a graceful fallback.  https://github.com/npm/arborist/pull/324

It's a good idea.  Nice find ;)

PR #322, opened 2021-09-17
Trying to fix Bug #321

### LINKED ISSUE #321: [BUG] Error when installing packages from local folder: "TypeError: Invalid comparator: file:../some-local/path"
# What / Why
Under the right circumstances it is impossible to install a package from a local folder (e.g. `npm i ../some-local/path`). The install command fails with the error message `TypeError: Invalid comparator: file:../some-local/path` (You can find the complete log file of the install below.

## How to reproduce?
I created a git repository which is setup to showcase the error. 

1. Clone the repo: https://github.com/sidney-pauly/minimal_ex_arborist_invalid_comp
2. Make sure you have the newest version of npm installed: `npm i npm@latest -g` (I'm not sure if this is a problem with older versions)
3. Go into the `/consuming-project` folder
4. Execute `npm i ../consumed-project`

Important conditions that need to be met to reproduce this issue:
The dependency needs to listed both in the devDependencies and the peerDependencies of the consuming package. Only then things break down. It might also be an issue if you try to do this with optionalDepndencies, but I did not try this yet.

### Current Behavior
The installation fails with the following error:
```
56 verbose stack TypeError: Invalid comparator: file:../consumed-project
56 verbose stack     at Comparator.parse (C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\semver\classes\comparator.js:37:13)
56 verbose stack     at new Comparator (C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\semver\classes\comparator.js:21:10)
56 verbose stack     at C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\semver\classes\range.js:121:20
56 verbose stack     at Array.map (<anonymous>)
56 verbose stack     at Range.parseRange (C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\semver\classes\range.js:121:8)
56 verbose stack     at C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\semver\classes\range.js:34:26
56 verbose stack     at Array.map (<anonymous>)
56 verbose stack     at new Range (C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\semver\classes\range.js:34:8)
56 verbose stack     at intersects (C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\semver\ranges\intersects.js:3:8)
56 verbose stack     at Arborist.[saveIdealTree] (C:\Users\sidne\AppData\Roaming\npm\node_modules\npm\node_modules\@npmcli\arborist\lib\arborist\reify.js:1211:62)
57 verbose cwd C:\Users\sidne\source\repos\sidney-pauly\minimal_ex_arborist_invalid_comp\consuming-project
58 verbose Windows_NT 10.0.19041
59 verbose argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Users\\sidne\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js" "i" "../consumed-project"
60 verbose node v16.6.0
61 verbose npm  v7.23.0
62 error Invalid comparator: file:../consumed-project
63 verbose exit 1
```

### Expected Behavior
The install should go through successfully and the local path should be put into the `package.json` instead of the version.

## Why don't you just remove the project from the devDependencies or peerDependencies?
In my case this leads to problems, because our team is using an automated system to create, bundle and package Typescript packages. This automated system is expected to work on existing projects without changing the package.json. As a result It breaks down because of this bug in arborist.

## How to fix?
I already looked into the code that causes the error. The problem seams to be that aborist tries to compare the previous version of the package with the current one. This fails because the current version is the local file path. See https://github.com/npm/arborist/blob/main/lib/arborist/reify.js (Lines 1208 to 1227). Maybe i will add a PR later

sidney-pauly:
Trying to fix Issue #321
To prevent the comparison of the relative file path to an actual version number I added a new variable which tracks if the current install is a local one.

If so no version comparisons are done

## References
Fixes #321

### DIFF
--- a/lib/arborist/reify.js
+++ b/lib/arborist/reify.js
@@ -1148,6 +1148,10 @@ module.exports = cls => class Reifier extends cls {
         }
 
         let newSpec
+        // True if the dependency is getting installed from a lcoal file path
+        // In this case it is not possible to do the normal version comparisons
+        // as the new version will be a file path
+        const isLocalDep = req.type === 'directory' || req.type === 'file'
         if (req.registry) {
           const version = child.version
           const prefixRange = version ? this[_savePrefix] + version : '*'
@@ -1179,7 +1183,7 @@ module.exports = cls => class Reifier extends cls {
           } else {
             newSpec = h.shortcut(opt)
           }
-        } else if (req.type === 'directory' || req.type === 'file') {
+        } else if (isLocalDep) {
           // save the relative path in package.json
           // Normally saveSpec is updated with the proper relative
           // path already, but it's possible to specify a full absolute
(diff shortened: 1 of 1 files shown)


isaacs:
The code looks fine, but this needs a test.

PR #323, opened 2021-09-20
feat: expose isDirect boolean for direct dependencies on the Vuln type

naugtur:
This is in relation to
npm/rfcs#372

I've extracted what can already be merged without much controversy from https://github.com/npm/arborist/pull/301
and provided test coverage.

@isaacs let me know if this is ok now, I'd like to merge this one and use it in npm-audit-resolver   
and then go back to working on types - the logic there demands I spend some more time figuring out what I want to achieve in more detail than I initially anticipated.

### DIFF
--- a/lib/vuln.js
+++ b/lib/vuln.js
@@ -82,6 +82,17 @@ class Vuln {
     }
   }
 
+  get isDirect () {
+    for (const node of this.nodes.values()) {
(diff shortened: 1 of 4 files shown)


isaacs:
Adjusted the code a bit so that workspace dependencies are also considered "direct", and replaced the flatmap with a for loop to exit a bit quicker.

PR #331, opened 2021-10-05
fix: drop node 10 support

wraithgar:
BREAKING CHANGE: this moves our `engines` support and testing matrix
forward to node12 LTS and above.


wraithgar:
pacote, run-script and metavuln calculator all need to be bumped here too to get on `node-gyp` 8

nlf (approved):
test failures look to be timeouts. might be time to raise the bar on the timeout a bit until we can figure out how to speed those tests up.

changes all look good, and i'm not concerned about the timeouts, so :shipit:

owen26:
I don't understand why were you dropping node v15 support in this PR. This is causing massive compatibility issues for libraries that depend on this lib. I can't find any clues or explanations after searching your issues board and changelogs.

ljharb:
I asume it’s because node 15 is EOL since April.

PR #332, opened 2021-10-05
chore: rename overridden prop to peerConflicted

nlf:
the new name is more accurate as to what it represents, and due to the
work on overrides the old name is going to become confusing. this is an
internal flag, so not considered a breaking change.

### DIFF
--- a/lib/arborist/build-ideal-tree.js
+++ b/lib/arborist/build-ideal-tree.js
@@ -880,7 +880,7 @@ This is a one-time fix-up, please be patient...
     const tasks = []
     const peerSource = this[_peerSetSource].get(node) || node
     for (const edge of this[_problemEdges](node)) {
-      if (edge.overridden) {
+      if (edge.peerConflicted) {
         continue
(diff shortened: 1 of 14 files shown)


wraithgar:
This may be an internal flag but it's not a symbol, the cli doesn't appear to use it but people *could* be?

Do we return edge objects from any arborist public methods?

nlf:
> This may be an internal flag but it's not a symbol, the cli doesn't appear to use it but people _could_ be?

_could_ be, true enough. it seems pretty unlikely, though. the flag isn't documented and previously had a very vague meaning..

> Do we return edge objects from any arborist public methods?

not directly, though you can certainly argue that since the `Edge` class is one of the public exports that we do. i don't have a particular concern with calling this a semver-major, so maybe it's best to just go ahead and do that and play it safe?

isaacs:
Sure, let's call it semver major.

isaacs (approved):
RSLGTM.

PR #336, opened 2021-10-14
chore: normalize ci workflow, raise tap timeout

nlf:
does what it says in the title, i want to see if these non-standard things that were present are actually necessary


nlf:
in general, i prefer not to turn off timeouts entirely for any reason. if we have a test that hangs i don't really want to find out after we've blown through hours/days of action compute time. it's not harmful in the grand scheme of things, but avoiding consuming resources we don't need to be consuming is a best practice i very much believe in.

PR #346, opened 2021-10-28
fix: don't compare spec for local dep vs existing

lukekarrys:
Fixes: #321 
Closes: #322


lib/arborist/reify.js
```diff
@@ -1173,6 +1173,10 @@ module.exports = cls => class Reifier extends cls {
         }
 
         let newSpec
+        // True if the dependency is getting installed from a lcoal file path
```
wraithgar, line 1176:
```suggestion
        // True if the dependency is getting installed from a local file path
```

PR #350, opened 2021-11-10
fix: ws no longer filter root deps

fritzy:
`arborist.excludeWorkspaceDependencySet` properly handles dependencies that
are both root and workspace deps for commands like prune, npm ls, etc.

Previously, commands like `arborist prune --workspacesEnabled=false` would prune packages erroneously if they were both a workspace and a root dependency.

### DIFF
--- a/lib/arborist/index.js
+++ b/lib/arborist/index.js
@@ -29,6 +29,7 @@
 const {resolve} = require('path')
 const {homedir} = require('os')
 const procLog = require('proc-log')
+const { depth } = require('treeverse')
 const { saveTypeMap } = require('../add-rm-pkg-deps.js')
 
 const mixins = [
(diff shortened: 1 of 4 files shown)


lib/arborist/index.js
```diff
@@ -122,28 +126,40 @@ class Arborist extends Base {
...
+      },
+      filter: node => node,
+      getChildren: (node, tree) =>
+        [...tree.edgesOut.values()].map(edge => edge.to),
+    })
+    return rootDepSet
+  }
 
-    return set
+  excludeWorkspacesDependencySet (tree) {
```
wraithgar, line 161: Why this extra function?

wraithgar (approved):
Just a question about an extra function, not a blocker. Looks good otherwise

lib/arborist/index.js
fritzy: Added as an alias, as I think the original name of the function could be misleading. It only makes sense if you're talking about it as a filter set, which is the intention, so I added an alias to what the set actually contains. It's very confusing. Maybe some comments would be enough.

PR #351, opened 2021-11-10
chore: convert to @npmcli/template-oss

lukekarrys:


.eslintrc.js
```diff
@@ -0,0 +1,14 @@
+// This file is automatically added by @npmcli/template-oss. Do not edit.
+
+const { readdirSync: readdir } = require('fs')
+
```
ljharb, line 4: maybe the autogenerated files should be marked in `.gitattributes` as such, so the diffs aren't shown?

wraithgar: That would be something to suggest in template-oss. I could see it being helpful to know what is being brought in though.

wraithgar (approved):

.eslintrc.js
ljharb: you can still expand the diffs and view them; it's more so that this PR's diff would by default only show the things that aren't auto-generated (like how package-lock.json diffs aren't expanded by default)

PR #354, opened 2021-11-22
feat: implement overrides

nlf:
This pull request implements overrides as described in [the spec](https://github.com/npm/rfcs/blob/main/accepted/0036-overrides.md)

~There are a couple of todo notes for myself in the code that I'm going through my notes to find context on, so this is a draft until I get those sorted out.~ I wasn't able to figure out what the notes were meant to address. Since tests tell a positive story here, I'm leaving the comments as they are in case of future issues and this is ready for review

### DIFF
--- a/lib/edge.js
+++ b/lib/edge.js
@@ -29,6 +29,7 @@ class ArboristEdge {}
 const printableEdge = (edge) => {
   const edgeFrom = edge.from && edge.from.location
   const edgeTo = edge.to && edge.to.location
+  const override = edge.overrides && edge.overrides.value
 
   return Object.assign(new ArboristEdge(), {
     name: edge.name,
@@ -38,12 +39,13 @@ const printableEdge = (edge) => {
     ...(edgeTo ? { to: edgeTo } : {}),
     ...(edge.error ? { error: edge.error } : {}),
     ...(edge.peerConflicted ? { peerConflicted: true } : {}),
+    ...(override ? { overridden: override } : {}),
   })
 }
 
 class Edge {
   constructor (options) {
-    const { type, name, spec, accept, from } = options
+    const { type, name, spec, accept, from, overrides } = options
 
     if (typeof spec !== 'string') {
       throw new TypeError('must provide string spec')
@@ -54,6 +56,10 @@ class Edge {
     }
 
     this[_spec] = spec
+
+    if (overrides !== undefined) {
+      this.overrides = overrides
+    }
 
     if (accept !== undefined) {
       if (typeof accept !== 'string') {
@@ -82,8 +88,11 @@ class Edge {
   }
 
   satisfiedBy (node) {
-    return node.name === this.name &&
-      depValid(node, this.spec, this.accept, this.from)
+    if (node.name !== this.name) {
+      return false
+    }
+
+    return depValid(node, this.spec, this.accept, this.from)
   }
 
   explain (seen = []) {
@@ -101,6 +110,10 @@ class Edge {
       type: this.type,
       name: this.name,
       spec: this.spec,
+      ...(this.rawSpec !== this.spec ? {
+        rawSpec: this.rawSpec,
+        overridden: true,
+      } : {}),
       ...(bundled ? { bundled } : {}),
       ...(error ? { error } : {}),
       ...(from ? { from: from.explain(null, seen) } : {}),
@@ -143,8 +156,13 @@ class Edge {
     return this[_name]
   }
 
-  get spec () {
+  get rawSpec () {
     return this[_spec]
+  }
+
+  get spec () {
+    return this.overrides && this.overrides.value ? this.overrides.value
+      : this[_spec]
   }
 
   get accept () {
@@ -213,6 +231,7 @@ class Edge {
     if (node.edgesOut.has(this.name)) {
       node.edgesOut.get(this.name).detach()
     }
+
     node.addEdgeOut(this)
     this.reload()
   }
--- a/lib/edge.js
+++ b/lib/edge.js
@@ -161,8 +161,24 @@ class Edge {
   }
 
   get spec () {
-    return this.overrides && this.overrides.value ? this.overrides.value
-      : this[_spec]
+    if (this.overrides && this.overrides.value && this.overrides.name === this.name) {
+      if (this.overrides.value.startsWith('$')) {
+        const ref = this.overrides.value.slice(1)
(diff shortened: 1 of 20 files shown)


test/arborist/build-ideal-tree.js
```diff
@@ -3110,3 +3163,522 @@ t.test('competing peerSets resolve in both root and workspace', t => {
...
+    t.equal(barEdge.valid, true, 'bar is valid')
+    t.equal(barEdge.to.version, '1.0.1', 'bar version is correct')
+
+    const bazEdge = barEdge.to.edgesOut.get('baz')
+    t.equal(bazEdge.valid, true, 'baz is valid')
+    t.equal(bazEdge.to.version, '1.0.1', 'baz version is correct')
+
+    const fooBazEdge = bazEdge.to.edgesOut.get('foo')
+    t.equal(fooBazEdge.valid, true, 'cyclical foo is valid')
+    t.equal(fooBazEdge.to.version, '2.0.0', 'override broke the cycle')
```
fritzy, line 3614: way to break the cycle, buddy!

test/arborist/build-ideal-tree.js
```diff
@@ -3110,3 +3163,522 @@ t.test('competing peerSets resolve in both root and workspace', t => {
...
+    t.equal(fooBarEdge.valid, true)
+    t.equal(fooBarEdge.to.version, '1.0.1')
+
+    const barEdge = tree.edgesOut.get('bar')
+    t.equal(barEdge.valid, true)
+    t.equal(barEdge.to.version, '2.0.0')
+  })
+
+  t.test('overrides a nested dependency that also exists as a direct dependency', async (t) => {
+    generateNocks(t, {
```
fritzy, line 3330: You use this same Nock fixture quite a bit. It could be done once and referenced.

lib/edge.js
```diff
@@ -101,6 +110,10 @@ class Edge {
       type: this.type,
       name: this.name,
       spec: this.spec,
+      ...(this.rawSpec !== this.spec ? {
+        rawSpec: this.rawSpec,
+        overridden: true,
+      } : {}),
```
ljharb, line 116:
```suggestion
      ...(this.rawSpec !== this.spec && {
        rawSpec: this.rawSpec,
        overridden: true,
      }),
```

lib/node.js
```diff
@@ -190,6 +193,17 @@ class Node {
...
 
+    if (overrides) {
+      this.overrides = overrides
+    } else if (loadOverrides) {
+      const overrides = this[_package].overrides || {}
+      if (Object.keys(overrides).length > 0) {
+        this.overrides = new OverrideSet({
+          overrides: this[_package].overrides,
+        })
+      }
```
ljharb, line 204: does this mean that `overrides` will not always be present? it seems easier to deal with if it's always an OverrideSet, just sometimes empty.

lib/edge.js
```diff
@@ -143,7 +156,28 @@ class Edge {
     return this[_name]
   }
 
+  get rawSpec () {
+    return this[_spec]
+  }
+
   get spec () {
+    if (this.overrides && this.overrides.value && this.overrides.name === this.name) {
+      if (this.overrides.value.startsWith('$')) {
```
ljharb, line 165: this seems like a strange sentinel/prefix approach. are you envisioning that an override can't ever start with `$`, even tho that might be a legitimate bundledDep folder name inside `node_modules`?

test/edge.js
```diff
@@ -1,5 +1,6 @@
 const util = require('util')
```
fritzy, line 1: This entire thing is just one test with a lot of assertions. Is this how we want to test data component classes?

lib/edge.js
nlf: [npm-package-arg](https://github.com/npm/npm-package-arg) defines `$` as an invalid character in a package name, anywhere we attempt to parse a spec will explode. i'd be surprised if it worked as a bundled dep name, have you seen it work that way?

i chose the `$` prefix specifically because npm-package-arg will throw when encountering it, but if that assumption doesn't hold water then we'll definitely have to come up with something else

lib/node.js
nlf: this was largely due to the nature of tests written in arborist. adding an empty OverrideSet to every node regardless of if overrides are in use breaks dozens of test snapshots. to avoid churn, i opted to omit the overrides unless they're in use. i agree it's weird though, and i do plan to remove that guard and redo the snapshots in a future pull request where it can be isolated

lib/edge.js
ljharb: nope! just thinking about possibilities.

Will every prior version of npa throw on it?

It's a fine assumption, I'm just not sure why a stringly typed interface is the ideal choice.

fritzy (approved):
This entire PR is ridiculously clean. Some of the testing is not ideal, but that's more of a reflection of what's already there to work with.

lib/edge.js
nlf: afaict all version of npa will throw on it.

using a string was a gut call. there are most definitely other approaches we could use here, but this was simple to parse and i knew it would be invalid as a "normal" package spec which is what i wanted.

lib/edge.js
```diff
@@ -38,12 +39,13 @@ const printableEdge = (edge) => {
     ...(edgeTo ? { to: edgeTo } : {}),
     ...(edge.error ? { error: edge.error } : {}),
     ...(edge.peerConflicted ? { peerConflicted: true } : {}),
+    ...(override ? { overridden: override } : {}),
```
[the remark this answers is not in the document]
nlf: good note, but yes unrelated. arborist in general could use some refactoring for things like this

test/arborist/build-ideal-tree.js
nlf: i liked keeping the calls to generateNocks inline in the test, it's handy to have the context of what a package's dependency tree looks like right in the test that exercises it. part of the struggles with the current testing that i experienced was having to go read fixtures to find out what the dep tree was even supposed to be.

it's a first pass at introducing nock here though, and i'd wager there's lots of improvements to be made

test/edge.js
nlf: lol it is not. i deemed it out of scope, however, to completely refactor the existing tests so i chose to add my tests in the same pattern as what existed and plan on a future refactor

Haprog:
Thanks a lot for implementing this!

PR #363, opened 2022-01-12
fix: _findMissingEdges missing dependency due to inconsistent path separators

salvadorj:
When installing a package with bundled dependencies with npm it can end up missing some dependencies that were bundled and attempt to fetch from the remote registry (see https://github.com/npm/cli/issues/2757). This issue only happens on Windows and it's caused by `_findMissingEdges` missing one of the bundled dependencies. The reason it only happens on Windows is because of this lookup in `this[_cache]` with a path that contains mixed path separators:
https://github.com/npm/arborist/blob/02f295f2d0caa6de2d3235c6eb1c180a8afa0f6f/lib/arborist/load-actual.js#L433-L435

On macOS we get a result back and avoid creating a new node. On Windows we get no result since the key in the cache is using backslashes but the lookup key has both backslashes and forward slashes. This results in a new node being created with the path of an existing node which in turn removes the existing node and its children from the tree. The children are however not loaded from the file system again because of this check:
https://github.com/npm/arborist/blob/02f295f2d0caa6de2d3235c6eb1c180a8afa0f6f/lib/arborist/load-actual.js#L353-L357

The end result is an unmet dependency and npm attempting to fetch it from the registry instead of using the bundled dependency.

I've changed the cache key by normalizing the path to avoid replacing existing nodes. I've setup a minimal test case where the difference in behavior (with and without normalization) shows up as an error on the outgoing edge of a node.

## References
Fixes https://github.com/npm/cli/issues/2757

### DIFF
--- a/lib/arborist/load-actual.js
+++ b/lib/arborist/load-actual.js
@@ -430,7 +430,7 @@ module.exports = cls => class ActualLoader extends cls {
           if (d.dummy) {
             // it's a placeholder, so likely would not have loaded this dep,
             // unless another dep in the tree also needs it.
-            const depPath = `${p}/node_modules/${name}`
+            const depPath = normalize(`${p}/node_modules/${name}`)
             const cached = this[_cache].get(depPath)
             if (!cached || cached.dummy) {
               depPromises.push(this[_loadFSNode]({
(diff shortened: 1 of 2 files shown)


fritzy:
@salvadorj This repo is being archived as arborist is now maintained as a workspace in the CLI. I recreated this PR as npm/cli#4261, and the commits maintain you as the author.

PR #364, opened 2022-01-19
Fix Tracker "idealTree:inflate:" already exists

kirtangajjar:
Fixes https://github.com/npm/arborist/issues/330 as per solution described in comments.


fritzy:
This PR has been migrated to https://github.com/npm/cli/pull/4298

